PR #7, opened 2026-06-15
Fix coach responsiveness, turn detection, and graceful stop

JINGBANZ:
## Summary

Fixes the four issues from the first live smoke run, plus the follow-ups found in subsequent live runs — all root-caused, then verified across **four multi-agent review rounds** (each finding adversarially verified before fixing). Up to date with `main` (merged PR #8).

## What was wrong → what changed

**1 & 3 — "Jarvis ignores me when addressed / stays silent."** The prompt was silent-by-default with no concept of direct address, so the model correctly chose silence. Now a hardened wake-word detector (`DirectAddress`, leading + trailing vocatives, filler-stripping) fires a `directAddress` trigger that bypasses the cooldown (honoring mute + its own ceiling); the prompt makes the model reply, with a spoken fallback so a direct address is never met with silence.

**2 — "Cuts me off mid-sentence."** Bare `server_vad` used the ~500 ms default. Verified `semantic_vad` is flaky in transcription mode, so instead **tuned `server_vad`** (`silence_duration_ms`, Config-driven) + a client-side debounce that coalesces fragments.

**4 — Noisy Stop.** Shutdown was already graceful; it just logged `1001` + POSIX 57 before checking `stopped`. Now gated, plus a fixed timer thread-affinity bug, reset reconnect counters, `.normalClosure`.

**Conversation quality (server-side state).** The brain never saw its own prior replies, so it couldn't follow up. Adopted the OpenAI **Conversations API** (one `conv_…` per session). After the first cut shipped four critical bugs (caught by review), reworked it to be stable: **index-based delta** (no clock-domain bug), **no forced tool** (prompt decides; fixes "Jarvis, check my screen"), **lazy tool-call close retained until provably sent** (no dangling/poisoned conversation), **coalesce-not-cancel** (nothing dropped), context **window when stateless**, and conversation-create **retry-after-backoff**.

**Disconnect / audio.** A drop no longer floods the log; mic audio is **buffered (60s cap) and replayed on reconnect** so a mid-sentence drop isn't lost.

**Observability.** `handleTrigger` returns a structured `TurnOutcome` and logs every "quiet" path (busy/coalesced, held-back, truncated, silent-by-model, cancelled, error), so the activity viewer always says *why*.

## Testing

- **88 tests pass** (was 36 at branch start), all added test-first; `JarvisApp` builds clean.
- Live-validated across three smoke runs: direct address, mid-sentence, graceful Stop, and the proactive screen-check all confirmed working.
- The one thing only a human run confirms is the live Conversations API continuity round-trip.


JINGBANZ:
## Multi-angle review + fixes (round 2)

Ran a 5-dimension review on the diff (correctness/concurrency, behavior-vs-claims, OpenAI/Apple API conformance, test quality, adversarial red-team), each finding adversarially verified before reporting. It confirmed **18 findings** and flagged the PR as *not mergeable as-is* — including two real bugs in the headline features. All are now fixed (latest commit), test-first.

### Must-fix (were real, now fixed)
- **Barge-in was defeated.** `cancelPrevious` cancelled the in-flight turn, but the replacement still hit the single-in-flight guard and was dropped `.busy` → fresh speech got **no reply** (worse than before). Fixed with a deterministic handoff: `TurnTaskBox` moved to JarvisCore; the replacement `await`s the cancelled predecessor so the slot frees first. New end-to-end test drives cancel-previous through `TurnTaskBox` + a real `CoachDriver`.
- **Wake-word missed trailing vocatives.** "Can you check this, Jarvis?" wasn't detected (anchor only scanned the first 3 tokens) — reproducing the exact symptom this PR fixes. `DirectAddress` now handles leading + trailing vocatives and disfluency fillers, while still rejecting discourse-marker narration ("So Jarvis told me…").

### Should-fix (fixed)
- `maxDirectAddressesPerMinute` was a **dead config knob** — now wired into `Guardrails`.
- A forced direct-address reply could still **truncate to silence** → added a spoken fallback (`.spokeFallback`); bumped `max_output_tokens` 400→768.
- Extracted the utterance coalescing into a testable `UtteranceBuffer` (the routing logic was untested in the app target).
- Gated `handle()`/`scheduleTurnDebounce()` on `stopped`; added an `.exhausted` test.

### Deferred (genuinely minor, tracked as fast-follows)
Test-hang-vs-fail hardening on the gated busy test; `ReconnectPolicy` extraction for backoff unit-testing; logging truncation even when a tool call is present.

**70 tests pass** (was 36 at branch start); `JarvisApp` builds clean. Live mic/websocket/model behavior still needs the human smoke run.

JINGBANZ:
## Review round 2 (conversation-state rework) + autonomous fixes

Ran a 5-dimension review on the rework (correctness/concurrency, conversation-state-vs-OpenAI-docs, behavior-vs-claims, test quality, red-team), each finding adversarially verified. It confirmed **21 findings — 4 were real bugs I'd introduced**, two of which re-created the "unanswered tool call poisons the conversation" failure the rework set out to kill. All fixed test-first (latest commit `5eb89c8`).

### Must-fix (were real, now fixed + regression-tested)
- **`capture_screen` on the final tool iteration dangled the conversation** — its result was never sent (loop cap), so every later turn 400s `No tool output found`. Now any unanswered capture is stashed and closed on the next turn.
- **The prior `speak` result was consumed before the send** — a `brainError`/cancel on the *next* turn stranded it and bricked the conversation. The close callId is now **retained until the send provably succeeds**.
- **Lost-trigger race** — "take next pending" and "release the slot" were separate critical sections, so a trigger arriving in the gap was orphaned. Replaced with atomic `claimOrPend` / `finishTurnOrTakeNext`.
- **Stateless fallback starved the model** — it sent only the per-turn delta (no server history), and one transient `createConversation` blip latched the whole session stateless. Stateless turns now send a context **window**; conversation creation **retries after a 60s backoff**.

### Should-fix (folded in)
Single-snapshot transcript delta (no duplicate-on-concurrent-append); audio flush drains in a loop before `connected` flips; new tests for the stateless path and directAddress coalesce priority; stale comments fixed.

### Deferred (genuinely minor)
A few coverage-hardening tests (advance-on-success on the capture path, mid-turn append, strengthening the coalesce assertion) and extracting the RealtimeTranscriber connected-flag state machine into a testable JarvisCore type.

**85 tests pass** (4 new regressions); `JarvisApp` builds clean. The live Conversations API round-trip still needs a human run — the continuity test ("Jarvis, I'm on two-sum" → then "what data structure?" without the name) is the key thing to verify.

PR #9, opened 2026-06-16
Overlay screen-capture invisibility: harden, verify, add regression tests

JINGBANZ:
## What & why

The overlay's "invisible during screen sharing" capability was **already implemented** via `NSWindow.sharingType = .none` — the same (and only) OS mechanism every alternative uses (Cluely, Interview Coder, Glass, cheating-daddy, Pluely, Natively; Electron's `setContentProtection(true)` and Tauri's `contentProtected: true` both map to it under the hood). This PR **verifies** it, **hardens** it, and adds **regression tests** — no new mechanism, because there isn't a better one.

### Empirical verification (macOS 26.5)

`.none` hides the overlay from every capture path tested — the `screencapture` CLI (Jarvis's own brain), `SCScreenshotManager`, and a **live `SCStream`** (the real Zoom/Meet/Teams path) — and it survives the API-key dialog's `NSApp` activation-policy flip. The widely-reported "macOS 15+ ScreenCaptureKit ignores `.none`" regression does **not** reproduce on Tahoe.

### Changes

- **Hardening:** re-assert `sharingType = .none` on every `show()` (defense-in-depth — Natively documents an activation-policy flip silently dropping the flag on some OS versions; it doesn't on 26.5, but the failure would be silent and high-impact).
- **Testability:** extract `OverlayPanel` into a `JarvisOverlay` library target (was in the `JarvisApp` executable; behavior unchanged).
- **Tests** — `Tests/JarvisOverlayTests/OverlayInvisibilityTests.swift`:
  - property tests (flag set at init + re-asserted on render) — run in CI, no permission needed;
  - opt-in ScreenCaptureKit capture test — `JARVIS_RUN_CAPTURE_TESTS=1` locally.
- **Docs:** new `wiki/overlay-invisibility.md` (mechanism, macOS-version nuance, limits, how to run the tests); comment clarifications in `OverlayPanel`/`ScreenCapture`.

## Test plan

- `./scripts/run-tests.sh` → `Test run with 91 tests in 14 suites passed` (incl. the 2 always-on property tests).
- `JARVIS_RUN_CAPTURE_TESTS=1 ./scripts/run-tests.sh --filter OverlayInvisibilityTests` (with Screen Recording granted) → all 3 pass; the capture test exercises real ScreenCaptureKit (~1.4s).
- `swift build` → `Build complete!`

## Limits / notes

`.none` is a best-effort window-server hint, not a security boundary: it does nothing against a phone camera, HDMI capture, OS-version drift, or proctoring **process** detection (a separate, stronger vector). The capture test can't run in hosted CI (no way to grant Screen Recording on a headless runner), so it's opt-in/local — researching whether CI can grant it; will follow up.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,13 +6,20 @@ let package = Package(
     platforms: [.macOS(.v14)],
     targets: [
         .target(name: "JarvisCore"),
+        // The AppKit overlay lives in its own library target (not the executable) so it can be
+        // imported by tests — see Tests/JarvisOverlayTests for the screen-capture invisibility checks.
+        .target(name: "JarvisOverlay", dependencies: ["JarvisCore"]),
         .executableTarget(
             name: "JarvisApp",
-            dependencies: ["JarvisCore"]
+            dependencies: ["JarvisCore", "JarvisOverlay"]
         ),
         .testTarget(
             name: "JarvisCoreTests",
             dependencies: ["JarvisCore"]
+        ),
+        .testTarget(
+            name: "JarvisOverlayTests",
+            dependencies: ["JarvisOverlay"]
         ),
     ]
 )
--- a/Sources/JarvisApp/AppDelegate.swift
+++ b/Sources/JarvisApp/AppDelegate.swift
@@ -1,5 +1,6 @@
 import AppKit
 import JarvisCore
+import JarvisOverlay
 
 @MainActor
 final class AppDelegate: NSObject, NSApplicationDelegate {
--- a/Sources/JarvisCore/ScreenCapture.swift
+++ b/Sources/JarvisCore/ScreenCapture.swift
@@ -6,8 +6,9 @@ public protocol ScreenCapturing: Sendable {
 }
 
 /// Uses the built-in `screencapture -x -t jpg` (silent). The main display only.
-/// NOTE: excluding Jarvis's own overlay window is handled by the overlay being a
-/// non-capturable panel (sharingType .none) in JarvisApp; see OverlayPanel.
+/// NOTE: excluding Jarvis's own overlay window is handled by the overlay being a non-capturable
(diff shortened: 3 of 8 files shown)


JINGBANZ:
### Follow-up: can the ScreenCaptureKit capture test run in CI? — Researched, answer is no (by design)

Decisive conclusion (cited research recorded in `wiki/overlay-invisibility.md` → *Why the capture test isn't in hosted CI*):

- **No public Apple API** grants Screen Recording (`kTCCServiceScreenCapture`) non-interactively — no entitlement, no test bypass. A PPPC/MDM profile can only *deny* Screen Recording, never silently allow it.
- **Hosted GitHub runners can't do it reliably.** Pre-granting via `sqlite3` on the system `TCC.db` only works because the macos-13/14 runner images happen to ship with **SIP disabled** (undocumented, removable). Even when granted, hosted runners return **blank/incomplete capture frames** — `actions/runner-images#8951` (open). The runners *do* have a window server, which is why our **property tests run fine in CI** (confirmed green here).
- **Every comparable project asserts the flag, not the pixels, in CI:** Electron added a private `isContentProtected()` getter specifically to test the flag without capturing — exactly what `overlaySetsCaptureExclusionAtInit` does. OBS and alt-tab-macos never pixel-test capture in CI either.

**So the split in this PR is the industry-standard approach:** flag/property tests gate CI; the pixel-level capture test is opt-in (`JARVIS_RUN_CAPTURE_TESTS=1`) and runs locally — verified passing on macOS 26.5 (~1.4s). If continuous pixel verification is ever required, the path is a **self-hosted Mac** with auto-login + a one-time manual grant, not the hosted-runner PR path.

Note: a capture-test *failure* on a future macOS would be a **true signal** (the OS stopped honoring `.none`, as reported on some macOS 15 builds) — not a flaky test.

PR #21, opened 2026-06-19
Open a fresh dev session on every Start

JINGBANZ:
## What

Every Start now opens a **fresh dev-mode session** (its own activity log, debug log, and screenshots) instead of resuming the previous run's.

## Why

A session was created **once per process launch** (`applicationDidFinishLaunching`). `start()`/`stop()` only built and tore down the coaching pipeline — they never rotated the session. So a **Stop → Start within one running process** kept appending to the same `jarvis-activity.jsonl`, which is the "it resumes the previous session" behavior reported.

## How

- **`AppDelegate`**: removed session creation from launch; added `beginNewSession()` (fresh `0700` dir + fresh `0600` `jarvis-debug.log` / `jarvis-activity.jsonl`), called from `start()` right after the `stop()` teardown. A session is now one coaching run (Start → Stop).
- **`ActivityViewer`**: `store` is mutable with a new `sessionDidChange(base:current:)` so the history browser re-points at the rotated session. Necessary because `ActivityLog.enable()` drops the prior observer — without it an *open* viewer would freeze on the old session's rows after a rotation.
- **wiki/build-and-run.md**: updated "per launch" → "per session / each Start".

## Behavior notes

- No session exists until the first Start, so launch-time messages go only to Console/NSLog (no coaching run = no session file).
- The just-finished session stays browsable and remains "current" until the next Start; `clearHistory` still spares the current one.

## Verification

- `swift build` clean; `./scripts/run-tests.sh` → 105/105 pass.
- Rotation logic lives in `JarvisApp` (intentionally not unit-tested per CLAUDE.md); live Stop→Start→new-log behavior to be confirmed via `./scripts/build-app.sh --dev`.

### DIFF
--- a/Sources/JarvisApp/App/AppDelegate.swift
+++ b/Sources/JarvisApp/App/AppDelegate.swift
@@ -57,7 +57,10 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         var sections: [SettingsSection] = [
             APIKeySection(keychain: keychain, onKeySaved: { [weak self] _ in
                 guard let self, self.transcriber != nil else { return }
-                self.menuBar.setRunning(self.start())
+                // Re-saving a key while running only re-applies it to the pipeline — it is NOT a new
+                // coaching run, so keep the current session (don't rotate logs). Session rotation is
+                // reserved for an explicit user Start.
+                self.menuBar.setRunning(self.start(freshSession: false))
             }),
             OverlaySection(appearance: appearance, applying: overlay),
         ]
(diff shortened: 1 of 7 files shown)


JINGBANZ:
## Multi-angle review + fixes

Ran a multi-agent review (4 lenses — correctness, concurrency, privacy, UX/lifecycle — with adversarial verification of every finding). 7 findings confirmed (deduping to 4 distinct issues), 3 correctly rejected (a non-deadlock, pre-existing unbounded-retention, and "disable-on-Stop" which is a design preference). Fixes applied in `44fd51a`:

1. **Cross-session log bleed (was the only issue contradicting the change's premise).** A turn cancelled by Stop could still emit after `start()` rotated the session — writing the previous run's screenshot/tip into the *new* session, plus a stale overlay tip. The detached screen capture doesn't inherit cancellation, so I guard the post-capture and pre-`speak` emit points on `Task.isCancelled`. Also closes a latent "speak after Stop". Covered by a new deterministic test (gated screen).

2. **Orphan empty sessions.** A Start that produced only lifecycle breadcrumbs (failed connect / instant Stop) cluttered the viewer's picker. `SessionStore.listSessions()` now hides past sessions with no coaching content; the current session is always shown. New test.

3. **Key-save coupling.** Saving an API key while running restarted via `start()` and so rotated the session — but re-applying a key isn't a new coaching run. `start(freshSession:)` keeps the current session for that path; rotation is reserved for an explicit user Start.

Left as intended: viewing-a-past-session gets refreshed to live on a new Start (matches "show it live").

107 tests pass (+2).

PR #22, opened 2026-06-19
Scale overlay line display time to length with an inter-line gap

JINGBANZ:
## What & why

The overlay showed every coaching line for a hard-coded 15s. This makes per-line display time **proportional to the line's length**, tuned to our actual situation rather than blindly copying the captioning spec.

### The model

```
per-line seconds = noticeBuffer + words × readingRate   (capped at max)
+ a brief blank gap between consecutive lines / tips
```

A deliberate hybrid:
- **Spine** from the captioning reading-speed standard (Netflix/BBC): duration ∝ length, with a floor and a cap.
- **Tuned** to us: the user is mid-conversation and only *glances* at the overlay (no audio echo), so the dominant cost is *noticing* the tip — a fixed `noticeBuffer` captions don't need, which also *is* the floor (so the old separate min knob is gone). The cap is tighter than film's ~7s because a coaching tip goes stale fast and an over-long one delays the next queued tip.
- **Borrowed** refinement: a brief blank **gap between lines** (the captioning minimum-gap rule) so a glancing eye registers that the text changed.

Resulting durations: 2-word ≈ 2.7s · 4-word ≈ 3.4s · 8-word ≈ 4.8s · 15-word ≈ 7.25s · long → 8s cap.

### Changes
- `OverlayTiming.swift` (new, Core/pure) — `displaySeconds(for:config:)`.
- `Config.swift` — `overlayNoticeBufferSeconds`, `overlaySecondsPerWord`, `overlayMaxDisplaySeconds`, static `overlayLineGapSeconds` (replaces `lineDisplaySeconds`/`overlayMin…`).
- `OverlayRendering` — `render` now takes a per-line `[TimeInterval]`; scalar convenience overload retained.
- `OverlayPanel` — plays each line for its own time and inserts the blank gap (`gapThenAdvance` + a self-enforcing `scheduleTick`). The tick scheduler is now **injectable** (`OverlayTickScheduler`); production keeps the real `DispatchOverlayScheduler`.
- `CoachDriver` — computes the per-line durations.
- `wiki/overlay-timing.md` (new) — full rationale + cited sources; linked from index, logged in status; records the deferred per-tip total-time cap.

### Review + hardening (later commits)
After a multi-angle review (concurrency / logic / design / tests / product), with each finding adversarially verified:
- Assert `CoachDriver` hands the overlay length-scaled per-line durations (was recorded, never checked).
- Gap test now asserts the panel stays **visible** during the blank; new test drives the `render([String], [TimeInterval])` empty-drop/alignment path directly.
- `scheduleTick` cancels any prior pending tick, so the "≤1 pending tick" invariant is self-enforcing.
- **De-flaked `overlayQueuesTipsInsteadOfInterrupting`**: it compared a 0.4s `Task.sleep` against a real 0.6s timer (200ms margin; `Task.sleep` only overshoots). Now driven by an injected `ManualOverlayScheduler` that fires ticks on demand — the ordering assertions no longer depend on any wall-clock relationship. (Mirrors the existing `Clock` injection in `JarvisCore`.)
- Deferred (your call): no per-*tip* total-time cap for now — validate pacing in the live smoke run first.

### Tests
`./scripts/run-tests.sh` → **113 pass**. Added `OverlayTimingTests` (Core), `overlayBlanksBetweenConsecutiveLines` + `overlayKeepsPerLineTimesAlignedWhenDroppingEmptyLines` (Overlay), and a CoachDriver duration assertion. De-flaked queue test verified stable across 5 consecutive runs.

> Verified by tests only — the "does a glancing user actually catch it / do multi-line tips linger" check is the live smoke run; the `Config` knobs are the tuning surface.


Sources/JarvisCore/Coach/CoachDriver.swift
```diff
@@ -280,7 +280,8 @@ public final class CoachDriver: @unchecked Sendable {
                 // the new session. This is the "speak after Stop" guard the TurnTaskBox relies on.
                 if Task.isCancelled { jlog("… turn cancelled (stopped) before speaking"); return .cancelled }
                 jlog("💬 \(lines.joined(separator: " "))")
-                overlay.render(lines, perLineSeconds: config.lineDisplaySeconds)
+                let perLine = lines.map { OverlayTiming.displaySeconds(for: $0, config: config) }
+                overlay.render(lines, perLineSeconds: perLine)
```
JINGBANZ, line 284: why are we hardcoding 0 here

Sources/JarvisOverlay/OverlayPanel.swift
```diff
@@ -188,3 +222,37 @@ public final class OverlayPanel: NSObject, OverlayRendering, OverlayAppearanceAp
...
+        DispatchQueue.main.asyncAfter(deadline: .now() + delay, execute: work)
+        return DispatchOverlayTickToken(work: work)
+    }
+}
+
+@MainActor
+private struct DispatchOverlayTickToken: OverlayTickToken {
+    let work: DispatchWorkItem
+    func cancel() { work.cancel() }
+}
```
JINGBANZ: Why do we need this here? We should not change our code because of the test. I believe tests should adapt to the code itself.

JINGBANZ: Agreed — reverted in 8f… (latest commit). Removed the `OverlayTickScheduler`/`OverlayTickToken` injection entirely and restored the direct `DispatchQueue.main.asyncAfter` timer. The flaky queue test now adapts to the real code instead: it gives the first line a long (3s) display window so the "still on screen when a newer tip arrives" check has a ~3s margin, and polls for the queued tip to take over rather than sleeping a fixed amount. No production change for the test; stable across repeated runs. (Kept only the `tickWorkItem?.cancel()` guard, which came from the review, not the test.)

Sources/JarvisCore/Coach/CoachDriver.swift
JINGBANZ: Want to make sure I fix the right thing — that line is `overlay.render(lines, perLineSeconds: perLine)`, where `perLine = lines.map { OverlayTiming.displaySeconds(for: $0, config: config) }`, so there's no literal `0` on it. The only hardcoded `0`s the PR introduced are `overlay.interLineGapSeconds = 0` in a few tests (to neutralize the inter-line gap so a timing assertion is clean). Did you mean those? If so I'll rework those tests to stop poking that value. Or if you're pointing at something else here, let me know what the `0` is and I'll address it.

PR #2, opened 2026-06-14
Phase 2: native Swift Jarvis LeetCode coach (Responses API + GA Realtime)

JINGBANZ:
## Summary

Builds **Phase 2** of Jarvis — the native macOS LeetCode-coach app — directly (Phase 1 fork PoC skipped). An always-on menu-bar app that transcribes you thinking aloud and, on each turn, calls `gpt-5.5` with `capture_screen` + `speak` tools to proactively render short coaching tips in an overlay.

Decisions captured in the wiki this round:
- **Skip Phase 1**, build the native Swift keeper directly.
- **Build in the main `forrest` account inside a git worktree** (restricted-account + App Sandbox hard-requirement waived for the personal build; hardened model kept as the future path).
- **Toolchain: SwiftPM + Command Line Tools — no full Xcode.** Manual `.app` bundle + ad-hoc `codesign`; macOS TCC prompts for Screen Recording + Microphone.

## Architecture

Two-target SwiftPM package:
- **`JarvisCore`** — pure, deterministic logic behind protocols (config, rolling timestamped transcript, guardrails, the coach tool-loop, the OpenAI client, sentence-splitting, screen-capture wrapper). Fully unit-tested with mocks; no GUI/permissions/network needed.
- **`JarvisApp`** — the side-effectful glue (NSPanel overlay, menu bar, AVFoundation mic, Realtime websocket). Compiles + launches.

## OpenAI integration — verified against live docs (2026-06)

- **Brain → Responses API** (`POST /v1/responses`): recommended for gpt-5.5 tool use. Flat function tools, system→`instructions`, `function_call`/`function_call_output` threading, `reasoning.effort=low`. `gpt-5.5` confirmed.
- **Transcription → `gpt-realtime-whisper`** (streaming STT) over the GA Realtime API: no `OpenAI-Beta` header, GA nested `session.audio.input` config, 24 kHz PCM, `server_vad` auto-commit, turn-end on `transcription.completed`. (The earlier `gpt-realtime-2` was not a real model ID.)
- **API key**: menu-bar paste → saves to Keychain → starts coaching immediately (no relaunch); `OPENAI_API_KEY` env is a headless fallback.

## Process

Brainstorm → plan → **fresh-agent review** (compile-tested on-machine; caught 3 blockers: no XCTest in CLT → swift-testing; Swift 6 strict concurrency; a Chat-Completions tool-loop bug — all fixed) → TDD build.

## Verification

[...]

PR #30, opened 2026-06-21
Reduce screen-question latency (~10s → ~3–4s)

JINGBANZ:
## Problem

A spoken question that needs the screen ("I fixed it — anything I can improve?") took ~10s to answer. The cause is structural: a screen question makes **two sequential, fully-awaited `gpt-5.5` calls** because the screenshot arrives via a `capture_screen` tool-call round-trip (the model decides it needs the screen, *then* a second call answers with the image), and nothing was streamed — so the overlay showed nothing until the second call fully completed. On top of that sits a ~1.4s VAD/debounce "are-you-done-talking" floor.

## Approach

Three constraints were kept deliberately: **`speak` stays a tool call**, **no screenshot is attached unless the model asks for it**, and **the model decides when it needs the screen**. So the decide-then-look round-trip stays; this PR cuts latency *inside* it:

1. **Capture overlaps the first brain call** (`CoachDriver`) — the screenshot is pre-warmed the moment a turn fires, concurrently with call #1, so it's already in hand when the model asks. Speculative: it's held in memory only and dropped (never written to disk) if the model never asks.
2. **The answer streams to the overlay** (`OpenAIBrainClient` + `SpeakLinesStreamParser`) — the brain call is read as an SSE stream (`stream:true`) and each `speak` line renders the instant it finishes generating, so first words appear ~1–2s into the answer instead of after the whole response. `speak` stays a tool call; the final `BrainResponse` is still decoded from the terminal event (byte-identical to the non-streamed result), and the overlay's queue/pacing is unchanged.
3. **A decisive capture decision** — the coach system prompt now tells the model to decide on the first turn whether it needs the screen and call `capture_screen` immediately, rather than answer-from-memory-then-look.

**Screenshot resolution is intentionally left at full `original`** — GPT-5.5 reads the shot at full resolution by default and the vision docs recommend `original` for dense images, so reliable code-reading outweighs the upload saving. (An earlier "lossless downscale" idea was dropped after verifying GPT-5.5's vision behavior against the official docs — it was based on the older GPT-4o rules and would have *reduced* what the model sees.)

## Outcome

- **Screen question:** ~10s → **~3.5–4.5s to first spoken word** (≈2–3s measured turn-end → first line; the rest is the kept VAD/debounce window).
- **Text-only question:** snappier, now streamed.

The remaining floor (two model passes + the VAD/debounce window) is structural and accepted; going lower would mean relaxing a kept constraint. See `wiki/architecture.md` §4 (Latency).

## Tests

**185/185 passing** across 29 suites (`./scripts/run-tests.sh`). New coverage: the incremental `SpeakLinesStreamParser` (chunk-boundary/escape/code cases), streaming `respond` over scripted SSE (line delivery + final-response equivalence + `incompleteReason`), streaming 429 retry incl. `Retry-After`, the parallel-capture overlap probe, and the live-vs-leftover overlay render (no double-render, no dropped line).

Built test-first via subagent-driven development with per-task review + a whole-branch review; the streaming retry path was made to honor `Retry-After` (shared with the non-streaming path) as a review fix.

## Still to verify live

`JarvisApp` isn't unit-tested. The latency win should be confirmed in the live smoke run — a "check my screen" question should stream its first line in ~2–3s after you stop talking.

### DIFF
--- a/Sources/JarvisCore/Coach/Brain.swift
+++ b/Sources/JarvisCore/Coach/Brain.swift
@@ -95,12 +95,23 @@ public protocol BrainClient: Sendable {
     /// only the NEW input each turn.
     func respond(messages: [ChatMessage], tools: [ToolDef], toolChoice: ToolChoice,
                  conversationId: String?) async throws -> BrainResponse
+    /// Streaming entrypoint: `onLine` receives each completed `speak` line as it generates. Has a
+    /// default (below) that ignores `onLine` and runs non-streaming, so conformers need not implement
+    /// it; `OpenAIBrainClient` overrides it with a real SSE stream.
+    func respond(messages: [ChatMessage], tools: [ToolDef], toolChoice: ToolChoice,
+                 conversationId: String?, onLine: (@Sendable (String) -> Void)?) async throws -> BrainResponse
     /// Create a server-side conversation and return its id. Default: a local stub (no server state),
     /// so mocks/tests need not implement it.
     func createConversation() async throws -> String
 }
 
 public extension BrainClient {
+    func respond(messages: [ChatMessage], tools: [ToolDef], toolChoice: ToolChoice,
+                 conversationId: String?, onLine: (@Sendable (String) -> Void)?) async throws -> BrainResponse {
+        try await respond(messages: messages, tools: tools, toolChoice: toolChoice,
+                          conversationId: conversationId)
+    }
+
     func createConversation() async throws -> String { "conv_local" }
 
     /// Convenience overloads so callers/tests need not pass every argument.
--- a/Sources/JarvisCore/Coach/CoachDriver.swift
+++ b/Sources/JarvisCore/Coach/CoachDriver.swift
@@ -202,13 +202,17 @@ public final class CoachDriver: @unchecked Sendable {
             \(ctx.promptLine)
             """))
 
-        // M1: pre-warm the screenshot NOW, concurrently with brain call #1, so when the model asks
-        // for the screen the image is already in hand (capture leaves the critical path). If the model
-        // never asks, the result is simply dropped. `screen.capture()` is synchronous/blocking and
-        // un-cancellable, so it runs detached off the cooperative pool — same as before, just earlier.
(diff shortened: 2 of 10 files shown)


JINGBANZ:
## Live test result — not shipping this PR

Ran the branch live (dev session `2026-06-21_21-42-20`). Two "look at my screen" questions; time from speech-end (`🗣 heard`) to the rendered hint (`💬`):

| Q | heard | 💭 thinking | 👁 looking (call #1) | 💬 hint (call #2) | total |
|---|-------|-------------|----------------------|-------------------|-------|
| 1 | 21:42:37 | 21:42:39 (+2s) | 21:42:42 (+3s) | 21:42:47 (+5s) | **~10s** |
| 2 | 21:43:10 | 21:43:10 (+0s) | 21:43:13 (+3s) | 21:43:19 (+6s) | **~9s** |

**Observed: the hint still appears ~10s after asking — no perceived latency improvement.**

Where the time goes, per the log:
- ~2s talk-detection (VAD silence + debounce) — kept by design
- **~3s for brain call #1** (the "do I need the screen?" decision)
- **~5–6s for brain call #2** (the answer with the screenshot)

The two sequential model calls — which this PR deliberately preserved (the model must decide it needs the screen, then answer with the image) — dominate the wall-clock. Pre-warming the screenshot only overlaps the ~0.3s capture; streaming the answer only helps *time-to-first-word*, which did not visibly materialize in testing. (The activity log also doesn't timestamp the first streamed line, so first-word latency can't be measured from it — a measurement gap.)

**Decision: not shipping.** The code is sound and well-tested (193 tests, two adversarial review rounds), but it does not meet the goal — perceived latency is still ~10s. Moving it for real requires relaxing a kept constraint: make the call-#1 decision much cheaper/faster (a small fast router model, or pre-attaching the screen for likely-screen questions and dropping the decision round-trip), and/or trimming the ~2s VAD/debounce floor. Those were deferred and are the actual levers.

Leaving the branch up for reference.

JINGBANZ:
Closing. It was tested live without a perceived latency gain, so it is not shipping (see the decision above).

PR #4, opened 2026-06-14
Menu-bar Start/Stop, no-auto-start, and a dev-mode live activity viewer

JINGBANZ:
## What & why

Active-development changes to make Jarvis testable by hand:

### Menu / lifecycle
- **Manual Start/Stop** in the menu bar; **Jarvis no longer auto-starts**. Pasting an API key stores it but does not begin coaching.
- Menu reduced to **two states**: ⚪️ stopped / 🟢 running. The user-facing **Mute** item is removed (the latent mute flag remains in `Guardrails` core).
- `AppDelegate` split into `start()` / `stop()` that build and tear down the audio + transcriber pipeline. Start cancels in-flight turns and reflects real connection state in the menu.

### Dev-mode live activity viewer
- New `JarvisCore/ActivityLog` mirrors **every `jlog` line** into a self-contained, auto-refreshing HTML page — watch Jarvis think in a browser without tailing a file.
- `CoachDriver`/transcriber log the decisions: 🗣 `heard: "…"` (your transcribed speech) · 🤫 silence · 💭 thinking · 👁 `capture_screen` · 💬 spoke · `… stayed silent`.
- Dev mode via the `--dev` launch flag; `scripts/run-dev.sh` rebuilds and launches via `open --args --dev --log-dir ./.jarvis`, auto-opening the viewer.
- **Logging is dev-only and owner-only:** outside `--dev`, only the unified log (no flat file); in dev, files are `0600`, truncated per session, in a **gitignored `.jarvis/`** — never world-readable `/tmp`.

### Docs
- Document the dev-mode viewer + privacy posture in README + `wiki/specification.md` + `wiki/sandbox.md`.
- Fix drift: ad-hoc → **stable `Jarvis Dev` signing identity** (permission persistence), no-auto-start + Start/Stop, mute removal, launch-via-`open` caveat, interjection counter (not "tokens/calls"), and the 26 → 36 test count.

## Verification
- `./scripts/run-tests.sh` → **36 tests pass** (incl. new `ActivityLogTests`).
- Built + stable-signed `Jarvis.app`; launched via `open`, confirmed permissions persist and the activity viewer streams live.
- Note: the no-key alert, menu state transitions, and the viewer's transcript line are verified by build + reasoning — still to be eyeballed in a live `run-dev.sh` smoke run.

## Review

[...]

PR #31, opened 2026-06-21
Add a global hotkey for an on-demand, one-trip screen hint

JINGBANZ:
## What & why

A screen-aware hint currently costs **two** brain round trips: the model emits a `capture_screen` tool call, the app screenshots and re-injects it, and the model reasons again. That second trip is pure latency, and it only happens reactively (on audio triggers).

This adds a deliberate, user-initiated path that does it in **one** trip. Pressing **⌥⌘J** while a session is running:
1. captures the screen **in the app**,
2. injects the image + a synthetic "give me a hint" user message into the **first** request, and
3. forces the `speak` tool, so a visible hint always comes back.

## Changes

**`JarvisCore` (testable logic):**
- `TriggerReason.manualHint` + its synthetic `promptLine` (`Triggers/Trigger.swift`).
- `CoachDriver.runTurn`: for `.manualHint`, captures the screenshot directly (off-pool, with the existing Stop-cancellation guard), appends it as a `.userImage`, and calls the brain with `toolChoice: .force("speak")`. All other behavior (conversation continuity, prior-close bundling, coalescing) is unchanged.
- `CoachDriver` also records the `.manualHint` trigger **and its pre-filled synthetic message** to the dev activity log, so the viewer shows exactly what the shortcut sent to the brain.

**`JarvisApp` (OS glue):**
- `Shortcuts/HotkeyController.swift`: registers **⌥⌘J** via Carbon `RegisterEventHotKey` — fires even when Jarvis isn't frontmost, **no Accessibility/TCC prompt**, no dependency.
- `AppDelegate`: arms a `requestManualHint` closure in `start()` / clears it in `stop()`, so the key works **only during a session** (beeps otherwise) and routes through `TurnTaskBox` for coalescing + Stop-cancellation.

## Design note: raw Carbon, not the KeyboardShortcuts package

The industry-standard `KeyboardShortcuts` package was the first choice, but **it can't build in this CLT-only environment** — every modern release uses SwiftUI macros (`@Entry` in 3.x, `#Preview` in 2.x) whose plugins ship only with full Xcode. Raw Carbon needs no macros, no dependency, and has the identical (no-permission) profile. The binding is fixed ⌥⌘J for now; a rebinding UI is deferred.

## Tests

- `Tests/JarvisCoreTests/Coach/CoachDriverManualHintTests.swift` (TDD): driver captures (not the brain), one brain call, forced `speak`, image present, `.spoke`; capture-failure and Stop-during-capture variants; plus an e2e test that the trigger + pre-filled message land in the activity log.
- Addressed a multi-angle review (logged hotkey-registration failures, added the cancellation + message-reaches-brain tests, fixed a stale doc comment), and merged latest `main` (overlay rename).
- Full suite: **182/182 green** (stable across repeated runs). App bundles + signs cleanly.
- Manual smoke (needs live permissions): press ⌥⌘J from another app → hint appears in one shot; press while stopped → beep.

Design recorded in the wiki: [`wiki/architecture.md` §2 On-demand hint](wiki/architecture.md#on-demand-hint-j) and the decision log in [`wiki/status.md`](wiki/status.md#key-decisions). (The intermediate design doc was folded in and removed.)


JINGBANZ:
## Multi-angle review — 6 dimensions, 3-vote adversarial verification

Reviewed across concurrency/Carbon FFI, CoachDriver logic, OpenAI request contract, app wiring/lifecycle, test adequacy, and style/reuse/privacy. Every finding was verified by 3 independent skeptics; only those ≥2/3 agreed were real are kept (11 of 14 survived; 3 were rejected).

**Verdict: Approve with minor changes.** No blockers, no data races, no contract bugs. The forced-speak + injected `input_image` request is valid (no 400 risk), the Carbon FFI has no race, and the start/stop closure invariant holds. The two "major" items are *missing tests*, not defects in shipping code.

> **Update:** findings #1–#4 below are now fixed in `3e96d20`.

### Should fix
- ✅ **[1] Silent hotkey-registration failure** — `HotkeyController.swift` *(flagged by 3 dimensions, 3/3 each)* — `InstallEventHandler`/`RegisterEventHotKey` discarded their `OSStatus`; a ⌥⌘J collision (`eventHotKeyExistsErr`) left the key silently dead. **Fixed:** both statuses now checked + `jlog`'d on failure.
- ✅ **[2] No test for the new manual-hint Stop-during-capture guard** *(major, 3/3)* — a separate path from the `capture_screen` guard, untested; a regression would stay green. **Fixed:** added `manualHintCancelDuringCaptureAbortsBeforeAnyBrainCall`.
- ✅ **[3] No test that the synthetic user message reaches the brain** *(major, 3/3)* — the "simulate a message as the user" half was unverified. **Fixed:** asserts the hint prompt + transcript land in the single request.
- ✅ **[4] Stale `ToolChoice.force` doc comment** — `Brain.swift` *(2 dimensions, 3/3)* — claimed force was unused by the coach. **Fixed.**

### Nice to have (not yet addressed)
- **[5] Manual hint can be silently dropped when busy** — `CoachDriver.swift:201` *(minor, 3/3)* — if a turn is in flight and a pending trigger exists, the keypress returns `.busy` and vanishes with no feedback. Consider letting `.manualHint` win the pending slot, or surface feedback on `.busy`.
- **[6] Extract the duplicated detached-capture idiom** — `CoachDriver.swift:201-207` *(minor, 3/3)* — shares with the `capture_screen` branch via a small `captureScreenshotOffPool()` helper.
- **[7] Untested: manual hint + stateful conversation with a prior pending close** *(minor, 3/3)*.
- **[8] `makeDriver` copy-pasted from `CoachDriverPipelineTests`** *(nit, 2/3)* — promote to a shared test helper.

*Rejected by adversarial verification (3 of 14): not shown.*

PR #33, opened 2026-06-22
Make the activity log a default in all modes, not just dev mode

JINGBANZ:
## What

The activity log — the in-app `WKWebView` viewer (Settings → **Activity**), its persisted `jarvis-activity.jsonl` + screenshots, and `jarvis-debug.log` — was gated behind the `--dev` launch flag. This un-gates it: the Activity tab, the live viewer, and per-session disk logging now run on **every** launch.

## Behavior

- `AppDelegate`: removed the `devMode` flag and its three gates (viewer construction, the Settings Activity tab, and `beginNewSession()`'s `guard devMode`). The viewer is always built; the tab is always present; each Start rotates a fresh session dir + logs. `devLogDirectory()` → `logDirectory()`.
- **Where logs land**: `~/Library/Caches/Jarvis/<session>/` by default; `build-app.sh --dev` now *only* routes them to the workspace `.jarvis/<session>/` (its sole remaining purpose — it no longer enables anything). Either way the dir is `0700`, files `0600`, fresh per session, never `/tmp`.

## Privacy posture (deliberate change)

This relaxes the prior **"nothing screen-/audio-derived is written to disk outside dev mode"** invariant, by explicit decision. The activity log (spoken tips, transcribed "heard:" lines, and the screenshots the model looked at) now persists **owner-only in all modes**. The raw mic audio and the live transcript are still **never** archived. `CLAUDE.md`, `README.md`, and the wiki (`sandbox`, `build-and-run`, `settings-window`, `architecture`, `status`, `index`) are updated to state the new posture; a dated decision is recorded in `wiki/status.md`.

## Verification

- `swift build` — clean.
- `./scripts/run-tests.sh` — **184 tests, all passing**. `ActivityLog`/`SessionStore` logic is unchanged (still gated by `enable()`); the un-gating lives in `JarvisApp`, which is intentionally not unit-tested. Worth a live smoke check that a normal launch now writes per-session dirs under `~/Library/Caches/Jarvis/`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,7 @@ and keep anything OS-bound thin and on the outside.**
 | `JarvisCore` | library | All the logic. **Foundation-only** — no AppKit / AVFoundation / ScreenCaptureKit / WebKit. Runs and is unit-tested on any machine. |
 | `JarvisOverlay` | library | The AppKit overlay (`NSPanel`, capture-invisibility). Its own target *so its behavior can be unit-tested* (`JarvisOverlayTests`) without dragging UIKit into Core. |
 | `CJarvisAEC` | C target | The acoustic-echo-cancellation edge: a pure-C facade over WebRTC AEC3. The C++ impl is prebuilt + statically merged (abseil included, zero runtime dylibs) into `lib/libjarvis-aec.a` by `scripts/build-aec.sh`, so `swift build` only links the archive — no C++ toolchain or vendored headers. Regenerate the `.a` only when bumping the webrtc version. |
-| `JarvisApp` | executable | The thin macOS shell: menu bar, capture, permissions, the dev viewer window. Wires Core + Overlay + AEC to the OS. Verified by a live run. |
+| `JarvisApp` | executable | The thin macOS shell: menu bar, capture, permissions, the activity viewer window. Wires Core + Overlay + AEC to the OS. Verified by a live run. |
 
 > **Put logic in `JarvisCore`.** If something can be written without an OS framework, it goes in Core
 > where a test can reach it. Reach outward to `JarvisOverlay` / `JarvisApp` only when you genuinely
@@ -30,8 +30,7 @@ and keep anything OS-bound thin and on the outside.**
 | Compile | `swift build` | Builds all targets. |
 | Test | `./scripts/run-tests.sh` | **Always use this, not raw `swift test`** — it fixes the swift-testing search path on CLT-only machines. Runs all three test targets. |
 | Build the app | `./scripts/build-app.sh [release\|debug]` | Bundles + signs `Jarvis.app` with the stable `Jarvis Dev` identity (so TCC grants persist). |
-| Run it | `./scripts/build-app.sh --run` | Build, then launch. Launch via `open`, never the bare binary. |
-| Dev mode | `./scripts/build-app.sh --dev` | Same as `--run`, but routes per-session logs to the workspace `.jarvis/` instead of `~/Caches/Jarvis`. The activity log is always on regardless. |
+| Run it | `./scripts/build-app.sh --run` | Build, then launch. Launch via `open`, never the bare binary. Per-session logs land in the workspace `.jarvis/`. |
 
 ## Layout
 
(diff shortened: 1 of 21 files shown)


JINGBANZ:
## Multi-angle review (5 lenses → adversarial verify → synthesize)

Ran an automated review across five lenses (correctness, privacy/security, disk lifecycle, docs consistency, simplification), with each finding adversarially re-verified against the code. **Verdict: sound — no blockers or high-severity issues.** 7 findings confirmed, 4 refuted.

**Privacy posture (the crux):** verified correct. The default base `~/Library/Caches/Jarvis/<session>/` is created `0700` (intermediates included) and all three artifacts (`jarvis-activity.jsonl`, `shot-N.jpg`, `jarvis-debug.log`) are `0600` — genuinely owner-only. No new persistence of raw mic audio or the live transcript; the transient `screencapture` temp file is still per-user and `defer`-deleted. Every prior statement of the old "no disk outside dev mode" invariant was rewritten to match.

**Fixed in `236507a`:**
- **Disk growth (the one real design gap):** the always-on log accumulated session dirs unboundedly. Added `SessionStore.pruneToMostRecent(_:)`, called at each Start to keep the **10 newest** (current always spared), + 2 unit tests.
- **`--dev` metadata leak (CWE-732):** `build-app.sh` now `chmod 700`s the workspace `.jarvis/` base so session-dir names aren't world-readable; fixed the comment that overstated it.
- **Dead optionality:** `activityViewer` is now non-optional (matching its siblings); dropped the always-true `if let` / optional-chain; `sections` is now `let`.
- **Stale docs/test:** removed remaining "dev mode / dev-mode viewer" wording (README ×2, wiki/CLAUDE.md, SessionStore header, a test fixture) and noted the 10-session retention in `sandbox.md`.

**Refuted (considered, don't apply):** "never /tmp" vs the `NSTemporaryDirectory()` fallback (that's the per-user `/var/folders/.../T` at 0700, not shared `/tmp`); in-session jsonl growth (pre-existing, `maxLines` is openly the in-memory cap); logs in purgeable `~/Library/Caches` (intentional — Apple's spot for regenerable data; Application Support would back sensitive content up to Time Machine/iCloud).

`swift build` clean; `./scripts/run-tests.sh` → **186 tests pass**.

JINGBANZ:
## Second multi-angle review (full PR, 5 lenses → adversarial verify → synthesize)

Re-ran the review over the **whole PR** (now covering `--dev` removal, the single workspace-`.jarvis/` location, the app-data fallback, pruning, and the README trim). **Verdict: merge with nits** — no blockers; the app-data fallback and pruning were both confirmed correct and safe. 5 findings confirmed, 3 refuted.

**Fixed in `839a813`:**
- **Medium — doc contradiction:** `wiki/architecture.md` §5 still said "Nothing is recorded to local disk in the MVP" (the one spot the invariant-rewrite missed). Now matches sandbox.md/CLAUDE.md.
- **Perms hardening:** `beginNewSession()` now best-effort `chmod 700`s the log base, so a *pre-existing* `0755` dir (e.g. an `Application Support/Jarvis` left by another tool) can't leak session-dir names — mirroring `FileSecretStore.setApiKey`. Also bound `logDirectory()` once (was recomputed 3×).
- **Retention scope clarified:** noted that the 10-session cap bounds session *count*, not per-session *size* (a single long run still grows its append-only logs/screenshots). Left as documented behavior rather than adding a byte-budget pruner (would be ceremony CLAUDE.md warns against).
- **Stale comments:** dropped four "dev"-mode comments describing the removed mode.

**Refuted (considered, don't apply):** the `secretFile.fileURL` coupling is deliberate and documented (logs intentionally sit alongside the key; avoids duplicating the Application Support derivation); the fallback can't create `Application Support` itself world-readable (standard 0700 per-user path); default-of-10 is a written-rationale judgment call, not a defect.

**Confirmed safe:** app-data fallback creates the full dir chain `0700` *before* any `0600` file — even with no saved key or an env-var key. Pruning orders newest-first by sortable id, `dropFirst(max(1,keep))` can't nuke the current session, symlinks skipped, bounded to session-shaped children — two passing tests.

`swift build` clean; `./scripts/run-tests.sh` → **186 tests pass**.

PR #35, opened 2026-06-25
Scale max_output_tokens with reasoning effort

JINGBANZ:
## Why

Setting reasoning effort to **high** truncated every coaching turn:

```
⚠️ response truncated (max_output_tokens) — not deliberate silence
```

Per OpenAI's [reasoning guide](https://developers.openai.com/api/docs/guides/reasoning), `max_output_tokens` is a **combined** cap over reasoning + visible output + formatting tokens — and they recommend reserving **≥25,000** for reasoning models. Our flat `768` default (tuned for `low`) meant high-effort reasoning exhausted the whole budget before the `speak` tool call was ever emitted, so the run came back `status:"incomplete"` / `reason:"max_output_tokens"` with no tip.

## What

- **Per-effort budget** (`ReasoningEffort.maxOutputTokens`) — scales with effort since the cap is almost entirely reasoning headroom (the tip itself is ~100–200 tokens):

  | Effort | Cap |
  |---|---|
  | none | 1,024 |
  | low | 2,048 |
  | medium | 8,192 |
  | high | 25,000 |

- **Wired through** `AppDelegate` so the cap tracks the effort dropdown.
- **Dropped the `768` trap default** in `OpenAIBrainClient` (now 2,048; real callers pass the per-effort value).
- **Per-turn usage logging** — `tokens — reasoning N, output M, cap C [incomplete]` — so the caps can be tuned *down* from real consumption, as the guide advises.

## Why not just remove the cap

Falling back to the context-window limit loses the safety valve against a single runaway reasoning pass spiking latency/cost on an always-on, every-turn assistant. Generous-but-bounded keeps both.

## Tests

`./scripts/run-tests.sh` → 195 pass. Added:
- `maxOutputTokensScalesWithEffort` — monotonic, high ≥ 25k, all > 768
- `encodesProvidedMaxOutputTokens` — the cap reaches the wire

Not yet verified live — needs a real run at high effort to confirm the truncation is gone (the new token logging will show actual reasoning spend).

### DIFF
--- a/Sources/JarvisCore/Coach/OpenAIBrainClient.swift
+++ b/Sources/JarvisCore/Coach/OpenAIBrainClient.swift
@@ -29,7 +29,7 @@ public struct OpenAIBrainClient: BrainClient, @unchecked Sendable {
                 timeout: TimeInterval = 15,
                 maxRetries: Int = 2,
                 backoffBaseSeconds: Double = 0.5,
-                maxOutputTokens: Int = 768,   // headroom so reasoning + a short tool call don't truncate
+                maxOutputTokens: Int = 2_048,   // combined reasoning+output cap; pass a per-effort budget (see ReasoningEffort.maxOutputTokens)
                 promptCacheKey: String = "jarvis-coach-v1",
                 send: Sender? = nil) {
         self.apiKey = apiKey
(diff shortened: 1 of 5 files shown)


Sources/JarvisCore/Coach/OpenAIBrainClient.swift
```diff
@@ -29,7 +29,7 @@ public struct OpenAIBrainClient: BrainClient, @unchecked Sendable {
                 timeout: TimeInterval = 15,
                 maxRetries: Int = 2,
                 backoffBaseSeconds: Double = 0.5,
-                maxOutputTokens: Int = 768,   // headroom so reasoning + a short tool call don't truncate
+                maxOutputTokens: Int = 2_048,   // combined reasoning+output cap; pass a per-effort budget (see ReasoningEffort.maxOutputTokens)
```
JINGBANZ: why are we hardcoding the token here, aren’t we setting this according to the reasoning effort?

JINGBANZ: Good catch — that bare default duplicated the per-effort budget and could drift from it (same hardcode class as the original 768 bug).

Fixed in 141aa8a: both `reasoningEffort` and `maxOutputTokens` defaults now derive from `ReasoningEffort.default`, so there's a single source of truth and no magic number. The real path is unchanged — `AppDelegate` passes `effort.maxOutputTokens` for the *selected* effort; the default just mirrors the default effort's budget as a fallback for tests.

PR #76, opened 2026-07-17
feat(screen): add stable screen change monitor

JINGBANZ:
> **Parked / draft.** This change is intentionally separated from #74 so the audio/transcription reliability fix can merge independently. Its base is `fix/interview-context-reliability`. Because this repository squash-merges, rebase the single screen commit onto updated `main` and retarget the PR after #74 merges.

## Problem

In the interview session that motivated #74, the last screenshot Jarvis sent to the coach showed a blank editor. The interviewer then posted the real question and the user began typing a solution, but those silent visual changes produced no transcript event. A screen-blind model repeatedly chose `stay_silent`, so it never saw the question or corrected the developing solution.

Prompt guidance alone improves capture decisions on speech turns, but it cannot discover a screen change when neither person says anything. Capturing every turn or every frame would create avoidable vision cost, CPU work, and privacy exposure.

## Solution

- Keep model-triggered `capture_screen` for ordinary speech, with explicit freshness guidance that a historical screenshot never proves the current screen state.
- Poll the active capture scope locally at low rate and bounded resolution using one-shot `SCScreenshotManager` captures. There is no persistent `SCStream`, so the monitor does not leave macOS's purple screen-sharing control visible.
- Compare downscaled visible pixels locally, use dirty-region metadata only as supporting evidence for tiny edits, and wait for quiescence before taking one full screenshot.
- Deduplicate full captures with normalized OCR plus a tolerant perceptual image hash, preserving meaningful non-text changes without reacting to small encoding drift.
- Hold the latest stable snapshot briefly so it can piggyback on the next speech request. If the interview remains silent, allow a screen-only fallback no more than once per minute.
- Record an explicit “stable screen change detected” activity row, persist the exact image before network send for audit, retry failed delivery safely, and prevent stale captures from acknowledging newer activity.
- Drop silent monitor-only turns from conversation history so repeated `stay_silent` checks do not inflate later prompts.

## Why this shape

This is a two-stage local gate: cheap one-shot pixels decide whether anything visibly changed; full capture, OCR, persistence, and model cost happen only after the new state settles. The model remains responsible for whether to speak, while the harness supplies fresh context when speech cannot reveal that a question or solution appeared.

The one-minute screen-only bound caps silent cost. Speech piggybacking remains immediate and does not add another brain request. Intermediate poll images stay memory-only and are never OCR'd, uploaded, or written to disk.

## Validation

- `swift build && ./scripts/run-tests.sh`
- 416 tests in 53 suites passed.
- Automated coverage includes visible-pixel stability under browser redraws, tiny dirty-region edits, OCR/perceptual deduplication, quiescence, candidate replacement, retry/re-arm behavior, capture ordering, piggyback timing, the screen-only request bound, audit-before-send, and silent-history removal.
- Manual session `2026-07-17_16-35-50_D4BF` exposed the browser-redraw quiescence issue and led to visible-pixel classification.
- Manual session `2026-07-17_19-39-22_38DE` exposed the persistent purple control and three short-interval brain calls; that led to one-shot polling plus the hard fallback interval.
- Final manual session `2026-07-17_20-18-16_621B` showed no persistent purple button, one accepted full screenshot, one corresponding brain request, and the explicit stable-screen activity marker.

## Known review findings before ready

The final combined-PR review produced three screen-specific findings after the last implementation commit. They are deliberately preserved here and must be addressed before this draft is marked ready:

- Prevent the asynchronous startup seed from baselining a screen update that appears after Start but before the first low-rate change report.
- Reject a full monitor capture if a newer screen change was observed while that capture was in flight, so stale pixels are never sent as the latest state.
- If startup seeding fails, keep the first dirty snapshot unbaselined until a brain request successfully consumes it, including displacement and settle-back cases.

## Review state

This PR is deliberately left as a draft for later consideration. #74 does not depend on it.

### DIFF
--- a/Sources/JarvisApp/App/AppDelegate.swift
+++ b/Sources/JarvisApp/App/AppDelegate.swift
@@ -319,12 +319,26 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         themTranscriber.onTerminalFailure = onThemTerminalFailure
 
         // One-clock capture + echo cancellation: a single aggregate device (mic + system tap) feeds
-        // the cleaned mic to the "me" socket and the raw system audio to the "them" socket, with AEC3
-        // run inside its IOProc. If the device can't be built, the whole capture is gone, so treat it
-        // as a full (mic-side) terminal failure.
+        // the cleaned mic to the "me" socket and the sample-preserving system timeline to the "them"
+        // socket, with AEC3 run inside its IOProc. If the device can't be built, the whole capture is
+        // gone, so treat it as a full (mic-side) terminal failure.
         let capture = AggregateEchoCapture(
-            onMicClean: { [weak transcriber] data in transcriber?.sendAudio(data) },
-            onSystem: { [weak themTranscriber] data in themTranscriber?.sendAudio(data) })
+            onMicCaptured: { [weak transcriber] sequence, samples, capturedAt in
+                transcriber?.recordCapturedAudio(
+                    sequenceNumber: sequence, sampleCount: samples, capturedAt: capturedAt)
+            },
+            onSystemCaptured: { [weak themTranscriber] sequence, samples, capturedAt in
+                themTranscriber?.recordCapturedAudio(
+                    sequenceNumber: sequence, sampleCount: samples, capturedAt: capturedAt)
+            },
+            onMicClean: { [weak transcriber] data, sequence, capturedAt in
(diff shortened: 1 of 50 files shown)


JINGBANZ:
Closing this parked draft. Main never adopted background screen polling, and the capture code it builds on has been rewritten.

PR #86, opened 2026-07-19
fix(capture): crop Chrome toolbar from screenshots

JINGBANZ:
## Summary

- crop Chrome and Chromium's standard tab strip and toolbar from active-window captures before both OCR and vision consume the JPEG
- resolve the selected window's owning bundle ID and convert the 82-point browser inset to backing-image pixels
- fail open to the original full-window image for unsupported apps, invalid geometry, or encoding failures
- document the zero-install behavior and cover bundle/scale/geometry decisions with Core tests

## Why

Chrome's browser-owned UI was consuming pixels and OCR tokens with unrelated tab titles, extension controls, and the address bar. Jarvis should focus on the visible page without requiring users to install a browser extension or start Chrome in a debugging mode.

## User impact

For standard macOS Chrome/Chromium windows, Jarvis now sends and OCRs the page content below the browser toolbar. Other applications, including Safari, remain unchanged.

## Validation

- `swift build && ./scripts/run-tests.sh` — passed, 426 tests across 54 suites
- `./scripts/run-tests.sh --filter BrowserChromeCropTests` — passed, 5 tests
- validated the supplied Chrome capture: 1937x1317 became 1937x1235, and OCR began at the LeetCode page UI instead of the tab/address bars

## Scope

This intentionally targets the standard 82-point Chrome/Chromium layout. Nonstandard layouts such as an additional bookmarks bar may leave browser UI visible; crop failures preserve the complete window rather than risk discarding page content.


JINGBANZ:
Closing. The capture code this patches was rewritten around Accessibility plus OCR text. A new issue is the place for it if the Chrome toolbar still matters.

PR #105, opened 2026-07-23
fix(brain): retry CLI subprocess timeouts in RetryingBrainClient

### LINKED ISSUE #104: CLI subprocess timeouts are never retried — domain mismatch defeats RetryingBrainClient
## Summary

A Claude Code / Codex coaching subprocess that hits its watchdog timeout is thrown as a timeout error, but `RetryingBrainClient` never retries it — the one automatic retry the wrapper exists to provide silently never fires. An identical OpenAI URLSession timeout *is* retried, so two timeouts the code goes out of its way to label with the same code behave differently.

## Root cause — producer/consumer value mismatch

**Producer** — `Sources/JarvisCore/Brain/AgentCLIProcessRunner.swift:143` throws:

```swift
throw NSError(domain: "AgentCLIProcessRunner", code: NSURLErrorTimedOut, userInfo: [...])
```

The code is deliberately `NSURLErrorTimedOut` (-1001) — a value only meaningful inside `NSURLErrorDomain`, whose only consumer in the repo is `RetryingBrainClient.isTransient`. That reuse signals clear intent for the timeout to be treated as transient.

**Consumer** — `Sources/JarvisCore/Brain/RetryingBrainClient.swift:28-42` only checks the timeout code when the *domain* matches:

```swift
if ns.domain == NSURLErrorDomain {
    return [NSURLErrorTimedOut, ...].contains(ns.code)   // gated on domain
}
if ns.domain == "OpenAIBrainClient" { return [408, 409, 500, 502, 503, 504].contains(ns.code) }
return false   // <-- CLI timeout (domain "AgentCLIProcessRunner") lands here
```

Because the thrown domain is `"AgentCLIProcessRunner"` — neither `NSURLErrorDomain` nor `"OpenAIBrainClient"` — `isTransient` returns `false`.

## Propagation is real (not dead code)

- `CLIBrainClient.respond` rethrows the runner error unchanged (`Sources/JarvisCore/Brain/CLIBrainClient.swift:79`).
- For CLI providers, `coachBase = CLIBrainClient(...)` and is wrapped in `RetryingBrainClient(base: coachBase)` exactly like the OpenAI client (`Sources/JarvisApp/App/AppDelegate.swift:227,263`).

## Failure scenario

Provider = Claude Code. A coaching turn's `claude -p` subprocess runs long and is SIGTERM'd at the watchdog timeout → `NSError(domain: "AgentCLIProcessRunner", code: -1001)` → `isTransient` returns `false` → the turn fails immediately with no retry. The equivalent OpenAI timeout (`NSURLErrorDomain` / `-1001`) is retried once. Retrying is explicitly documented as safe here: a lost response "cannot duplicate a screenshot or spoken tip" (`RetryingBrainClient` doc comment).

## Files involved

- `Sources/JarvisCore/Brain/AgentCLIProcessRunner.swift:143` (throw site)
- `Sources/JarvisCore/Brain/RetryingBrainClient.swift:25-43` (`isTransient`)
- `Tests/JarvisCoreTests/Brain/RetryingBrainClientTests.swift` (missing case that hides the gap — it covers `URLError(.timedOut)` and `NSError("OpenAIBrainClient", 503)` but no `AgentCLIProcessRunner` timeout)

## Suggested approach

Pick one (minimal, one-line each):

1. In `RetryingBrainClient.isTransient`, add:
   ```swift
   if ns.domain == "AgentCLIProcessRunner" { return ns.code == NSURLErrorTimedOut }
   ```
2. Or change the producer to throw with `domain: NSURLErrorDomain` so the existing `NSURLErrorTimedOut` check catches it.

Option 1 keeps the CLI runner's distinct domain (useful for diagnostics) and localizes the policy in the one place that decides retryability — preferred.

## Acceptance criteria

- A CLI subprocess timeout (`NSError(domain: "AgentCLIProcessRunner", code: NSURLErrorTimedOut)`) is retried exactly once by `RetryingBrainClient`.
- A new test in `RetryingBrainClientTests.swift` scripts that error → success and asserts `callCount == 2` (mirroring `retriesOneTransientFailure`). It fails before the fix and passes after.
- Non-timeout `AgentCLIProcessRunner` errors (if any other code is ever thrown) remain non-transient.
- `swift build && ./scripts/run-tests.sh` is green.

JINGBANZ:
## Summary

`AgentCLIProcessRunner` throws its watchdog timeout as `NSError(domain: "AgentCLIProcessRunner", code: NSURLErrorTimedOut)` — deliberately reusing the URL-loading timeout code to signal transient intent. But `RetryingBrainClient.isTransient` only honored `NSURLErrorTimedOut` under `NSURLErrorDomain`, so the runner's own domain fell through to the non-transient default. The result: an OpenAI URLSession timeout was retried once, while an identical CLI (Claude Code / Codex) subprocess timeout failed immediately with no retry — the one automatic retry the wrapper exists to provide silently never fired.

## Change

- `RetryingBrainClient.isTransient` — add an `AgentCLIProcessRunner` domain branch that treats `NSURLErrorTimedOut` as transient, keeping the runner's distinct domain for diagnostics (option 1 from the issue). Any other code from that domain stays non-transient.
- `RetryingBrainClientTests` — add `retriesCLISubprocessTimeout` (CLI timeout → success, asserts `callCount == 2`, mirroring `retriesOneTransientFailure`) and `doesNotRetryNonTimeoutCLIError` (a non-timeout code from that domain is not retried).

## Testing

The full `swift build && ./scripts/run-tests.sh` gate can't run on this Linux CI runner — the package links `AppKit`/`WebKit` via the macOS-only overlay/viewer/app targets, and `swift test` compiles the whole tree. The affected code lives entirely in the Foundation-only `JarvisCore` library, which builds cleanly here (`swift build --target JarvisCore` ✅). The gate should be run on macOS to confirm the new tests pass.

Fixes #104

### DIFF
--- a/Sources/JarvisCore/Brain/RetryingBrainClient.swift
+++ b/Sources/JarvisCore/Brain/RetryingBrainClient.swift
@@ -39,6 +39,11 @@ public struct RetryingBrainClient: BrainClient, Sendable {
         if ns.domain == "OpenAIBrainClient" {
             return [408, 409, 500, 502, 503, 504].contains(ns.code)
         }
+        // The CLI runner labels its watchdog timeout with `NSURLErrorTimedOut` to signal the same
+        // transient intent as an OpenAI URLSession timeout, but keeps its own domain for diagnostics.
+        if ns.domain == "AgentCLIProcessRunner" {
+            return ns.code == NSURLErrorTimedOut
+        }
         return false
     }
 }
--- a/Tests/JarvisCoreTests/Brain/RetryingBrainClientTests.swift
+++ b/Tests/JarvisCoreTests/Brain/RetryingBrainClientTests.swift
@@ -41,6 +41,30 @@ import Testing
         #expect(base.callCount == 2)
     }
 
+    @Test func retriesCLISubprocessTimeout() async throws {
+        let base = RetryScriptBrain(script: [
+            .failure(NSError(domain: "AgentCLIProcessRunner", code: NSURLErrorTimedOut)),
+            .success(.init(toolCalls: [.staySilent(callId: "ok")])),
+        ])
+        let client = RetryingBrainClient(base: base)
+
+        _ = try await client.respond(messages: [.user("hi")], tools: coachTools)
+
+        #expect(base.callCount == 2)
+    }
+
+    @Test func doesNotRetryNonTimeoutCLIError() async {
+        let base = RetryScriptBrain(script: [
+            .failure(NSError(domain: "AgentCLIProcessRunner", code: 1)),
+        ])
+        let client = RetryingBrainClient(base: base)
+
+        await #expect(throws: (any Error).self) {
+            _ = try await client.respond(messages: [.user("hi")], tools: coachTools)
+        }
+        #expect(base.callCount == 1)
+    }
+
     @Test func doesNotRetryPermanentFailure() async {
         let error = NSError(domain: "OpenAIBrainClient", code: 401)
         let base = RetryScriptBrain(script: [.failure(error)])


JINGBANZ:
Closing. RetryingBrainClient no longer exists and #335 removed CLI coaching, so #104 is obsolete.

PR #126, opened 2026-08-05
build(transcription): restore pre-macOS 26 SDK build compatibility for Apple Speech

### LINKED ISSUE #125: Restore pre-macOS 26 SDK build compatibility for Apple Speech
## Summary

PR #123 adds opt-in Apple Speech transcription using the macOS 26 `SpeechAnalyzer` APIs. The implementation works with the macOS 26 SDK, but the repository's macOS 15 CI runner cannot compile source files that name those newer API types.

This should be handled as a focused compatibility change without changing OpenAI transcription behavior or the macOS 26 Apple Speech data path.

## Evidence

- Related PR: #123
- Failing workflow: https://github.com/JINGBANZ/jarvis/actions/runs/30985871135
- CI environment: macOS 15, Swift 6.1.2
- Representative compiler errors:
  - `cannot find 'SpeechTranscriber' in scope`
  - `cannot find type 'SpeechAnalyzer' in scope`
  - `cannot find type 'AnalyzerInput' in scope`
  - `cannot find 'AssetInventory' in scope`
- The same branch builds on the macOS 26.5 SDK with Swift 6.3.3, and the serialized local suite passes 605 tests across 71 suites.

## Root cause

`@available(macOS 26.0, *)` protects runtime execution, but it does not hide unknown type names from an older SDK/compiler. Every direct reference to the new Speech framework APIs therefore needs a compile-time SDK/toolchain boundary.

Affected areas currently include:

- `AppleSpeechTranscriber`
- `AppleSpeechModelPreparation`
- Apple Speech locale discovery and availability in `APIKeyControls`
- Apple Speech construction in `TranscriptionSessionFactory`

## Proposed implementation

- Compile the real `SpeechAnalyzer` adapter only when the build toolchain exposes the macOS 26 Speech APIs.
- Keep the existing runtime macOS 26 availability checks.
- On older SDKs, compile a small fallback that:
  - reports Apple Speech as unavailable;
  - disables its Settings option and locale discovery;
  - prevents session construction through the existing typed user-facing failure path.
- Keep the shared `TranscriptionSession`, audio-format contract, coaching coordinator, and all OpenAI providers unchanged.
- Add a way to force-build the fallback path with the newer SDK, if practical, so it does not silently bit-rot.

A candidate boundary is a compiler/SDK conditional such as `#if compiler(>=6.2) && canImport(Speech)`, but the implementation should verify the earliest toolchain that actually contains these symbols rather than assuming the version indefinitely.

## Acceptance criteria

- The existing macOS 15 GitHub Actions job compiles and passes.
- A macOS 26 build still exposes and runs the Apple Speech provider.
- On an older SDK/toolchain, Apple Speech is unavailable without affecting OpenAI transcription.
- No macOS 26-only Speech symbol is parsed on the fallback compilation path.
- `swift build` passes.
- `SWT_EXPERIMENTAL_MAXIMUM_PARALLELIZATION_WIDTH=1 ./scripts/run-tests.sh` passes.
- The change lands separately from the provider/refactor implementation.

JINGBANZ:
## Summary

Restores compilation of the Apple Speech transcription work (PR #123) on the repository's **macOS 15 / Swift 6.1.x** GitHub Actions runner, without changing OpenAI transcription behavior or the macOS 26 Apple Speech data path.

`@available(macOS 26.0, *)` gates *runtime* execution but does not hide the new `SpeechAnalyzer` / `SpeechTranscriber` / `AnalyzerInput` / `AssetInventory` type names from an older SDK/compiler, so every direct reference needs a *compile-time* boundary.

### What changed
- **`Package.swift`** — single source of truth: defines `APPLE_SPEECH_SDK` for the `JarvisApp` target only when `compiler(>=6.2)` (Xcode 26 is the first SDK to ship these symbols and pairs with Swift 6.2). The manifest is built by the same toolchain that builds the package, so this reflects the real build compiler.
- **`AppleSpeechTranscriber.swift`** — the whole `SpeechAnalyzer`-backed adapter is wrapped in `#if APPLE_SPEECH_SDK`.
- **`AppleSpeechModelPreparation.swift`** — type and `prepare(localeIdentifier:)` signature exist on both paths (so `AppDelegate` links unchanged); only the Speech body is gated. On older SDKs it throws the typed unavailable failure.
- **`TranscriptionSessionFactory.swift`** — Apple Speech construction gated; the fallback case is unreachable because preparation already fails first.
- **`APIKeyControls.swift`** — Settings locale discovery + availability gated; on older SDKs the locale option shows "Unavailable" and is disabled.

The shared `TranscriptionSession` contract, audio-format contract, coaching coordinator, and all OpenAI providers are untouched.

### Force-build the fallback (anti-bit-rot)
`JARVIS_FORCE_APPLE_SPEECH_FALLBACK=1 swift build` (from a clean `.build`) compiles the fallback even on a new toolchain.

### Verification
- This fix targets `feat/apple-speech-transcription` (PR #123) because the Apple Speech code it guards lives only on that branch, not on `main` — the issue asks for the compatibility change to land separately from the provider implementation.
- Built on the Linux CI worker (Swift **6.1.3**, i.e. `compiler(<6.2)` → the fallback path, matching the macOS 15 CI condition): the manifest parses, `JarvisCore` builds, and a `JarvisApp` build surfaces **zero** Speech-symbol errors (only `AppKit`/`AVFoundation` module-missing errors inherent to Linux). Full `swift build` + `SWT_EXPERIMENTAL_MAXIMUM_PARALLELIZATION_WIDTH=1 ./scripts/run-tests.sh` verification must run on the macOS CI (both the macOS 15 fallback job and a macOS 26 real-path job), which this branch can't execute on a Linux worker.

Fixes #125

Generated with [Claude Code](https://claude.ai/code)


JINGBANZ:
Closing. #128 made this fix and #125 is closed.

PR #158, opened 2026-08-12
fix(build): separate development app identity

JINGBANZ:
## Summary

- build local development bundles as `Jarvis Dev.app` with bundle id `com.jarvis.coach.dev`
- keep production packaging unchanged as `Jarvis.app` / `com.jarvis.coach`
- make the transcription benchmark target the development bundle
- add a Gate guard and document independent TCC, preferences, migration, and coexistence behavior

## Why

The self-signed development build and Developer ID release previously used the same app name and bundle id but had incompatible designated requirements. macOS could therefore show an enabled “Jarvis” privacy row for one identity while repeatedly prompting the other. Giving development and production distinct application identities makes their TCC records and Launch Services registrations unambiguous and durable.

## Validation

- `swift build && ./scripts/run-tests.sh` — passed; 755 tests in 89 suites
- `./scripts/build-app.sh debug` — built and signed `Jarvis Dev.app`
- built development plist — `Jarvis Dev` / `com.jarvis.coach.dev`
- installed production plist — `Jarvis` / `com.jarvis.coach`
- host `codesign --verify --deep --strict` and designated-requirement checks passed for both bundles
- XcodeBuildMCP Swift package debug build passed

Live launch/capture was intentionally not performed, so the first `Jarvis Dev` TCC prompt still needs manual confirmation on the target Mac.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,10 @@ cd jarvis
 ./scripts/build-app.sh --run
 ```
 
-On the first local build, macOS asks whether `codesign` can use the `Jarvis Dev` key; choose
-**Always Allow**. The downloaded release does not use this local development identity.
+The script creates and opens `Jarvis Dev.app`, whose name, bundle id, preferences, and macOS privacy
+grants are independent from an installed `Jarvis.app` release. On the first local build, macOS asks
+whether `codesign` can use the `Jarvis Dev` key; choose **Always Allow**. macOS then asks once for the
+development app's own capture permissions. The downloaded release does not use this local identity.
 
 For signing details, permission recovery, and other commands, see
 [Build and run](./wiki/build-and-run.md). Always launch the app through the build script or `open`,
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -54,8 +54,10 @@ cd jarvis
 ./scripts/build-app.sh --run
 ```
 
-On the first local build, macOS asks whether `codesign` can use the `Jarvis Dev` key; choose
-**Always Allow**. The downloaded release does not use this local development identity.
+The script creates and opens `Jarvis Dev.app`, whose name, bundle id, preferences, and macOS privacy
+grants are independent from an installed `Jarvis.app` release. On the first local build, macOS asks
+whether `codesign` can use the `Jarvis Dev` key; choose **Always Allow**. macOS then asks once for the
+development app's own capture permissions. The downloaded release does not use this local identity.
```
JINGBANZ: can we keep this concise then it will launch the dev version?

PR #164, opened 2026-08-16
feat(release): distribute jarvis as a dmg

JINGBANZ:
## Summary

- replace the ZIP and separate checksum uploads with one stable, labelled `Jarvis.dmg` app asset
- place `Jarvis.app` beside an `/Applications` shortcut so installation is an in-window Finder drag
- sign, notarize, staple, mount, and inspect the final DMG before a draft release can be published
- link directly to the latest DMG from the README and add tag-specific install guidance to releases
- record the distribution decision and keep the release configuration guarded by the normal test gate

## User experience

The user opens `Jarvis.dmg`, drags `Jarvis.app` onto the **Applications** shortcut in that Finder
window, ejects the Jarvis disk image, and opens Jarvis from Applications. GitHub continues to add its
automatic source archives, but the project publishes only one app artifact and the primary download
link bypasses the Assets list.

This changes future releases; it does not mutate the already-public v0.1.3 assets.

## Verification

- `swift build && ./scripts/run-tests.sh` — passed; 754 tests in 89 suites
- native DMG smoke — created, verified, mounted read-only, found exactly two root entries, confirmed
  `Applications -> /Applications`, and detached successfully
- release shell syntax, release configuration guard, app identity guard, workflow YAML parsing, and
  `git diff --check` — passed

The actual Developer ID signature and Apple notarization require the protected release credentials,
so they were not invoked locally. The release job performs those steps and now refuses publication
unless the final DMG passes its signature, notarization-ticket, mounted-layout, SDK, architecture,
license, `syspolicy_check`, and Gatekeeper checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,14 +20,11 @@ session for an immediate, screen-aware hint.
 
 [**Download Jarvis for Apple silicon (macOS 14.2+)**](https://github.com/JINGBANZ/jarvis/releases/latest/download/Jarvis.dmg)
 
-1. Open `Jarvis.dmg`.
-2. In the Finder window, drag `Jarvis.app` onto the **Applications** shortcut.
-3. Eject the Jarvis disk image, then open Jarvis from **Applications**. Jarvis appears in the menu
-   bar rather than the Dock.
-4. Allow Microphone access and, for screen-aware hints, Screen Recording when macOS prompts.
-5. From the menu-bar icon, open **Settings**, choose transcription and brain providers, add any
-   credential those providers require, then select **Start Jarvis**. Allow System Audio Recording
-   when Start requests it so Jarvis can hear the other side.
+After installing, open Jarvis from **Applications**. It appears in the menu bar rather than the Dock.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -18,13 +18,14 @@ session for an immediate, screen-aware hint.
...
-2. Move `Jarvis.app` to **Applications**, then open it. Jarvis appears in the menu bar rather than
-   the Dock.
-3. Allow Microphone access and, for screen-aware hints, Screen Recording when macOS prompts.
-4. From the menu-bar icon, open **Settings**, choose transcription and brain providers, add any
+[**Download Jarvis for Apple silicon (macOS 14.2+)**](https://github.com/JINGBANZ/jarvis/releases/latest/download/Jarvis.dmg)
+
+1. Open `Jarvis.dmg`.
+2. In the Finder window, drag `Jarvis.app` onto the **Applications** shortcut.
+3. Eject the Jarvis disk image, then open Jarvis from **Applications**. Jarvis appears in the menu
+   bar rather than the Dock.
```
JINGBANZ: this should not be needed because the instruction should be self explanatory

Issue #219, opened 2026-08-29
feat(brain): index user-provided prep material for retrieval-augmented coaching


EdHasNoLife:
## Motivation

Users preparing for interviews often maintain personal notes: a set of system-design (or behavioral/coding) topics with the answers they've already worked out. Right now Jarvis has no way to use that material — it only ever sees the live transcript and, on demand, the screen. Letting the coach draw on prepared answers when a live question resembles a prepped topic should produce noticeably better, more personalized coaching.

## Design

Fits the existing on-demand, model-invoked tool pattern (`capture_screen`/`speak`/`stay_silent`) rather than always-in-context injection, since the corpus can be large (a personal knowledge base, not a few Q&A pairs) and always injecting it would tax every request regardless of relevance.

- **Input**: a new Settings section where the user points at one or more local files/folders (markdown/text) containing their prep notes. Prep material is persistent user config (like `Config`/`Secrets`), not session-scoped `.jarvis/` data — the user maintains it externally (Obsidian, plain notes, past interview writeups); Jarvis only reads it.
- **Indexing**: at Session Start, Jarvis chunks the configured material (by heading/section) and builds a local lexical index — BM25 or equivalent keyword ranking, Foundation-only, no network call, no new external dependency. This was chosen over embeddings-based semantic search specifically to stay provider-neutral: an embeddings step (OpenAI or otherwise) would require an API key even when the brain route is CLI-only (Claude Code/Codex), breaking the existing "CLI-only route needs no API key" guarantee. Semantic search can be revisited later if BM25 proves too weak in practice (see Non-goals).
- **Tool**: a new `search_prep_notes(query)` tool joins `[capture_screen, speak, stay_silent]`. The model calls it when a live question resembles a prepped topic; it returns the top-k matching chunks (with source) as a tool result, mirroring how `capture_screen` already works.
- **Prompt**: a short addition to the coach system prompt instructing the model to prefer the substance/structure of a strong prepared-answer match, without quoting it verbatim when it doesn't fit the live phrasing.

## Expected behavior

- User configures prep material paths in Settings; it gets indexed locally at Session Start.
- A live question matching an indexed topic triggers a visible `search_prep_notes` call in Activity, and the reply reflects the prepped content.

## Failure / degradation behavior

- No material configured → the tool is absent from the tool set entirely (zero cost, matches how screen capture already degrades when unavailable).
[...]

PR #220, opened 2026-08-29
feat(prep): add local prep-material source configuration to Settings

EdHasNoLife:
## Summary
- Slice 1 of #219: a new "Prep Material" Settings tab lets the user point Jarvis at local files/folders of interview notes (text/Markdown/PDF/Word) it can later index and reference during coaching.
- `PrepMaterialSource`/`PrepMaterialPreferences` (JarvisCore, Foundation-only) store only the path, never a copy of file contents — mirrors `BrainPreferences`' fallback-target persistence pattern.
- Sources are added via a native `NSOpenPanel` (files + folders, filtered to `.txt`/`.md`/`.pdf`/`.docx`), removable individually, and flagged with a warning when a path no longer resolves on disk.
- Nothing is copied or uploaded: persistence uses this macOS account's own UserDefaults domain, the same as every other setting.
- Out of scope here: indexing/parsing file content and the `search_prep_notes` coaching tool — a follow-up slice.

## Test plan
- [x] `swift build` — clean
- [x] `./scripts/run-tests.sh` — 791/791 tests pass, including new `PrepMaterialPreferencesTests` (round-trip, dedup-on-add, remove-by-id, malformed-entry handling); ghost-mode presentation-API gate passes
- [x] Live-verified in `Jarvis Dev.app`: opened Settings → Prep Material, added a PDF and a Markdown file via the picker, confirmed both list correctly with working remove buttons and the file-type filter applied

### DIFF
--- a/Sources/JarvisApp/App/AppDelegate.swift
+++ b/Sources/JarvisApp/App/AppDelegate.swift
@@ -28,6 +28,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private let brainPreferences = BrainPreferences()
     private let transcriptionPreferences = TranscriptionPreferences()
     private let screenPreferences = ScreenCapturePreferences()
+    private let prepMaterialPreferences = PrepMaterialPreferences()
     private var activityViewer: ActivityViewer!    // embedded as the Settings Activity tab
     /// Two provider sessions feeding one shared transcript: mic → `.me`, system audio → `.them`.
     private var transcriber: (any TranscriptionSession)?       // "me" (mic)
@@ -176,6 +177,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             connectionsSection,
             OverlaySection(appearance: appearance, caption: overlayCaption, box: overlayBox),
             DisplaySection(preferences: screenPreferences),
+            PrepMaterialSection(preferences: prepMaterialPreferences),
             ActivitySection(viewer: activityViewer),
         ]
         settingsWindow = SettingsWindow(sections: sections)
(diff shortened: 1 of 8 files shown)


EdHasNoLife:
Ran a code review pass on this PR and addressed the findings:

- **Fixed:** per-source `exists()` checks were running synchronously on the main thread on every tab open/add/remove — a stat() on a slow network volume or sleeping external disk could freeze the whole Settings window. Moved off-main via `Task.detached`, with a generation guard so a stale check can't clobber a newer render.
- **Fixed:** multi-selecting N files triggered N sequential UserDefaults read+write round-trips. Added a batched `PrepMaterialPreferences.add([PrepMaterialSource])` so one multi-select add is one read + one write.
- **Fixed:** `ClosureButton` was copy-pasted from `BrainTargetRowView.swift` instead of shared — extracted to `Sources/JarvisApp/Settings/ClosureButton.swift`.
- **Fixed:** a doc comment on `PrepMaterialPreferences` claimed Jarvis "reads the live file at Session Start," which isn't implemented yet (that's the follow-up indexing slice) — corrected the wording.
- **Not fixed (deliberate):** dedup-on-add compares raw path strings with no case/symlink normalization. A correct fix needs a synchronous filesystem call (resolving symlinks or querying canonical case), which would reintroduce the same main-thread-blocking risk just fixed above — and the failure mode is a harmless duplicate row, not data loss. Leaving as-is.

All 791 tests + the ghost-mode gate pass; re-verified live in the running dev app.

JINGBANZ (approved):
lgtm

Issue #225, opened 2026-08-29
feat(overlay): show a retrieved diagram/image in the Overlay Box


EdHasNoLife:
## Dependency

**Blocked on #226** (adds `search_prep_notes`) merging into `main` first — this issue's entire mechanism is "when that tool retrieves a low-text chunk, attach the image to that same turn," so there's nothing to hook into until it lands. Not blocked on #224 (see below).

## Context

Follow-up from #219/#224. Live-testing the prep-notes search feature surfaced a case where the coach's text-only reply lost real fidelity a source diagram had (see #224: an image-based diagram in a PDF extracts no text, and the reply reconstructed an approximate, incorrectly-ordered pipeline from surrounding prose instead of the diagram itself).

Even once #224's OCR fallback lands, OCR only recovers flat text from a diagram — it does not preserve the diagram's visual structure (box layout, arrows, grouping, direction). Producing something that preserves that structure needs actual diagram *interpretation*, not just text recognition.

## Scope for this slice (revised)

Show a **Mermaid-syntax text description** of the diagram in the **Overlay Box only** — not a rendered picture, and not the Overlay Caption.

- **Text, not a rendered graphic.** The Overlay Box already renders text (it's how spoken tips show today) — displaying Mermaid syntax as text needs zero new rendering capability. Actually rendering it as a visual diagram was considered and rejected for this slice: the Box is a pure native AppKit panel today with no web-rendering engine, and embedding one (e.g. mermaid.js via a WebView) is a real architecture change, not a content-format change. Revisit only if plain Mermaid text proves insufficiently readable in practice.
- **No separate vision/LLM call at index time.** Generating correct Mermaid syntax requires interpreting the diagram's layout — something flat OCR cannot do (it recognizes text runs, not box/arrow topology). Doing that interpretation as a dedicated step in `PrepMaterialIndexBuilder` would need its own vision-capable model call on every Session Start for every diagram-containing page (no caching, by design), and would force prep-material indexing — which today deliberately needs no network call, no API key, and works identically regardless of brain provider (see `PrepMaterialIndex`'s doc comment, and the #219 decision log) — to depend on whichever brain happens to be configured. That's a real invariant, not an incidental detail.
[...]

Issue #224, opened 2026-08-29
feat(prep): OCR fallback for image-only pages when extracting prep-material PDFs


EdHasNoLife:
## Context

Follow-up from #219 (prep-material search). Discovered during live testing of the indexing slice (PR from `feat/prep-notes-search`, stacked on #220).

`PrepMaterialIndexBuilder.extractPDFText` (Sources/JarvisApp/PrepMaterial/PrepMaterialIndexBuilder.swift) uses `PDFKit`'s `PDFPage.string` to extract text per page. This only reads real text layers — it returns nothing for a page whose content is a rendered image (e.g. a diagram exported from Excalidraw/diagrams.net and embedded as a raster image).

## Evidence

Live-tested against a real user PDF ("Design a Top-K URL System for LinkedIn Posts"): the document's High-Level Design section is a diagram image with box labels and an explicit sequencing annotation ("partitioned by article_id — canonicalize BEFORE partitioning"). Verified directly with a standalone extraction script against all 8 pages: none of the diagram's text (box labels, the ordering annotation) appears anywhere in the extracted output. Page 2's extracted text ends right at the "High Level Design" heading with nothing after it.

Consequence observed live: `search_prep_notes` correctly retrieved the relevant chunks (retrieval/BM25/tool-loop all worked as designed), but since the diagram's exact flow and ordering were never in the indexed text, the coaching reply reconstructed an approximate — and in this case incorrectly ordered — pipeline from the surrounding prose instead of the ground-truth diagram.

## Proposed fix

Jarvis already has OCR infrastructure for exactly this shape of problem: `ScreenTextRecognizer` (Sources/JarvisApp/Capture/) uses Apple's Vision framework (`VNRecognizeTextRequest`) to OCR live screen captures for `capture_screen`. Reuse the same technique as a fallback in `PrepMaterialIndexBuilder`: when a PDF page's `PDFPage.string` yields little or no text, render that page to an image (`PDFPage.thumbnail(of:for:)` or similar) and run it through the same Vision-based OCR path, folding the recognized text into that page's extracted content.

## Non-goals

- No OCR for `.docx`/`.txt`/`.md` — only PDF pages that already yield near-empty text extraction.
- No layout/diagram-structure understanding — OCR gives flat recognized text, same fidelity as the existing screen-capture OCR sidecar (may garble token order for a complex diagram; still strictly better than zero text).
[...]

Issue #48, opened 2026-07-06
ActivityLog.cssClass missing ⌨️ case — manual-hint entries render unstyled in activity viewer


JINGBANZ:
## Context

When the user presses the hint hotkey, `CoachDriver.runTurn` logs a line prefixed with `⌨️`:

```swift
// Sources/JarvisCore/Coach/CoachDriver.swift:167
if reason == .manualHint { jlog("⌨️ hint shortcut — \(ctx.promptLine)") }
```

`ActivityLog.cssClass(for:)` maps the leading emoji of each log line to a CSS class that colors it in the activity viewer:

```swift
// Sources/JarvisCore/Diagnostics/ActivityLog.swift:168–174
static func cssClass(for message: String) -> String {
    let m = message.trimmingCharacters(in: .whitespaces)
    if m.hasPrefix("💬") { return "say" }   // green
    if m.hasPrefix("👁") { return "see" }   // yellow
    if m.hasPrefix("🗣") || m.hasPrefix("🤫") { return "hear" }  // blue
    if m.hasPrefix("💭") || m.hasPrefix("…") { return "think" }  // grey
    // ...error fallback, then "" default
}
```

There is **no case for `⌨️`**, so the manual-hint log line falls through to `""` — the same empty class used for lifecycle noise like `"Jarvis: coaching started."`. In the activity viewer the hint-shortcut invocation renders plain white, visually indistinguishable from unimportant lifecycle breadcrumbs, even though it represents a user-initiated turn trigger just like `🗣` (voice turn end) and `🤫` (silence).

The existing `ActivityLogTests.cssClassKeysOnLeadingMarker` test covers all six currently-handled prefixes but does not include `⌨️`, so the gap is unpinned.

## Files involved

| File | Location |
|---|---|
| `Sources/JarvisCore/Diagnostics/ActivityLog.swift` | `cssClass(for:)`, ~line 172 |
| `Tests/JarvisCoreTests/Diagnostics/ActivityLogTests.swift` | `cssClassKeysOnLeadingMarker`, ~line 6 |

## Acceptance criteria

1. `ActivityLog.cssClass(for: "⌨️ hint shortcut — Trigger: …")` returns `"hear"` (the class shared by `🗣` and `🤫` — all three are user-triggered turn events).
2. `ActivityLogTests.cssClassKeysOnLeadingMarker` includes an assertion that pins this behavior.
3. `./scripts/run-tests.sh` passes.

## Suggested approach

[...]

PR #226, opened 2026-08-29
feat(prep): add prep-notes retrieval to the coaching tool loop

EdHasNoLife:
## Summary

Completes the core of #219: the coach can now actually draw on the user's configured prep material during a live session, on top of #220's source configuration (merged).

- `Sources/JarvisCore/PrepMaterial/` — `PrepMaterialSearching` port, chunk model, and a hand-rolled Okapi BM25 index. Pure, Foundation-only, no dependency — works the same regardless of brain provider (settled in #219: no semantic/embedding search, no external API for retrieval). Added to `scripts/check-coaching-kernel.sh`'s covered paths, same shape as `Screen/`.
- `Sources/JarvisApp/PrepMaterial/PrepMaterialIndexBuilder.swift` — the extraction adapter behind that port: plain text/Markdown read directly, PDF via `PDFKit`, Word via the stock `textutil` CLI (same "borrow the OS tool" pattern as `screencapture`). Each file extracts on its own detached task.
- `CoachDriver`/`CoachAttemptRunner` — `search_prep_notes` joins `[capture_screen, speak, stay_silent]` as a non-terminal tool call, offered only once a source actually produced usable text. The port installs asynchronously via `CoachDriver.installPrepMaterial(_:)` (building the index is real file I/O, so it can't block Session Start), then is snapshotted into each attempt's `AttemptBrain` at selection time — the same discipline `plan` and `configuredRoute` already follow post-#218.
- New `ActivityEvent.prepNotesSearched` (with its own `cssClass` mapping, deliberately not repeating #48's "missing emoji case renders unstyled" bug) and a `searchPrepNotesContinuation` `RequestPhase` for audit provenance.
- One coach-prompt tweak: "Your only actions are capture_screen, speak, and stay_silent" is now "Your only actions are the tools available to you" — the enumeration was already going stale for any session where the fourth tool isn't offered, with zero behavioral change either way.

Note: this branch was originally built against the pre-#218 `CoachDriver` (before it split into a scheduler + `CoachAttemptRunner`) and has since been fully re-integrated against the merged lean-coaching-core refactor — the diff above reflects the current architecture, not a stale one.

## Live verification

Tested end to end in the running dev app against two real prep-material files (a PDF and a Markdown cheatsheet):
- Spoke a system-design question matching the PDF's topic; `search_prep_notes` fired (visible in Activity as `📎 checked prep notes for "..." — found 3 matches`), retrieved 2 chunks from the PDF and 1 from the Markdown file (confirmed via the session's `brain-traffic.jsonl`), and the reply drew on them.
- Surfaced a real extraction-fidelity gap in the process: an image-based diagram in the PDF (an Excalidraw-style export) extracts **zero** text via `PDFKit` — confirmed directly against all 8 pages of the real file. The model's reply reordered a pipeline step because the diagram's explicit sequencing annotation was never in the indexed text; retrieval and the tool loop themselves worked correctly. Filed as a follow-up rather than addressed here: #224 (OCR fallback reusing the existing `ScreenTextRecognizer` Vision-based path for image-heavy PDF pages). A related follow-up, #225, tracks showing such a diagram image directly in the Overlay Box.

## Test plan

- [x] `swift build` — clean
- [x] `./scripts/run-tests.sh` — 840/840 pass (24 exercise this feature directly: BM25 index, chunker, full `CoachDriverPrepMaterialTests` tool-loop pipeline, Activity event rendering), ghost-mode and coaching-kernel dependency guards both pass
- [x] Live-verified in `Jarvis Dev.app` per above


EdHasNoLife:
Ran a code review pass (round 1) and addressed the findings:

- **Fixed:** docx extraction produced one giant, unsplittable chunk per file — `textutil` separates paragraphs with a single newline (verified against a real Word-style RTF with ordinary paragraph breaks), not the blank-line convention `PrepMaterialChunker` splits on. Normalized each non-empty line to its own paragraph before chunking; confirmed empirically that this restores correct per-paragraph splitting.
- **Fixed:** a `Process` pipe deadlock in docx extraction — `standardError` was an undrained `Pipe()` while the same call blocks on `stdout`'s `readDataToEndOfFile()`; enough stderr output from `textutil` would hang forever. Routed stderr to `/dev/null`, matching existing precedent (`SyntheticSpeechFixtures.swift`) for the same hazard.
- **Fixed:** the index-build `Task` wasn't tracked or cancelled on Stop, unlike history compaction — an untracked task kept reading files/spawning `textutil` after session teardown and kept the old `CoachDriver` alive via a strong capture. Now tracked, cancelled on Stop, and stops dispatching new per-file work once cancelled.
- **Fixed:** `PrepMaterialChunk.swift` and `PrepMaterialSearching.swift` each bundled two public types, violating `AGENTS.md`'s one-type-per-file rule and this code's own claimed `Screen/` precedent. Split into four files.
- **Fixed:** `search_prep_notes` results weren't carried forward via `work.observations` the way `capture_screen`'s are, so a failed follow-up request in the same attempt forced a full re-search on retry. Now mirrors that pattern.

All 840 tests + both static guards (ghost-mode, coaching-kernel) pass. Round 2 review in progress.

EdHasNoLife:
Round 2 review, addressed:

- **Fixed:** a stale doc comment referencing `CoachDriver.effectiveTools`, which doesn't exist post-lean-coaching-core-split.
- **Fixed:** `search_prep_notes` called with no prep material configured now fails the attempt instead of silently returning an empty result with a real-looking "checked prep notes" Activity entry. A schema-enforced provider can't hit this (the tool's absent from the request entirely), but a CLI protocol reconstructs calls from free-form prompt text with no such guarantee.
- **Fixed:** argument parsing is now lenient (`JSONSerialization`, not a strict `[String:String]` Decodable dictionary) so an unexpected sibling field from a non-schema-enforced provider can't fail the whole call.
- **Fixed:** capped concurrent file extraction at 8, so a large prep-material folder can't spawn an unbounded number of simultaneous `textutil`/PDFKit calls.
- **Fixed:** factored the `capture_screen`/`search_prep_notes` tool-continuation boilerplate into one shared helper — worth doing now since #224/#225 suggest more tool variants are coming to this exact spot.
- **Fixed:** `.isRegularFileKey` was fetched but never used — now actually filters out directories/bundles that happen to match a supported extension.
- **Fixed:** `PrepMaterialIndex`'s three parallel arrays replaced with one array of a per-chunk struct, removing a desync risk.
- **Fixed:** added the missing `@unchecked Sendable` justification comment on the test fake.
- **Documented, not further engineered:** the docx paragraph heuristic can't distinguish a manual line break from a real paragraph break — inherent to plain-text extraction, bounded impact, not worth deeper heuristics right now.
- **Not done, deliberately:** generalizing tool-set construction into a reusable "optional tool with availability predicate" mechanism — speculative with only one gated tool existing today.

All 844 tests + both static guards pass, including new coverage for the rejected-call path and the lenient-parsing fix.

EdHasNoLife:
One more fix, found while explaining the docx normalization to the user and double-checking whether PDF had the same issue: it does. Verified directly against the real 8-page PDF used in earlier live testing — \`PDFPage.string\` never emits a blank-line boundary either, identical to \`textutil\`'s docx output. The prior fix only normalized the docx path, so a PDF page dense enough to lack a natural page-boundary break would still extract as one oversized, unsplittable chunk (this particular document's pages happened to be small enough to dodge it — confirmed the whole 8-page doc reduced to only 8 paragraph units before the fix).

Applied the identical normalization to \`extractPDFText\`, factored into one shared helper so both format converters can't drift apart on the same fix again. Confirmed with the same PDF: 227 paragraph units after the fix, vs. 8 before.

844/844 tests + both guards still pass.

JINGBANZ:
## Review

Checked locally on `c66f565`: build clean, 844/844 tests pass. The red CI check is an unrelated flaky timing test (`DiagnosticEvidenceTests` mailbox overflow) — it passes here.

Design looks right. Two bugs, and two things worth settling.

### 1. Bug: the prep-notes result wipes out the screenshot

`CoachAttemptRunner.swift:439` replaces the observations instead of adding to them:

```swift
work.observations = [.user(JarvisPrompts.Coach.prepNotesResult(results))]
```

If the model captures the screen, then searches prep notes, then the next request fails, the retry has lost the screenshot and has to capture again. `PendingCoachingWork`'s own comment says the latest screen observation is kept — this breaks that. Append instead of replace.

### 2. Bug: Codex is told the wrong list of tools

`JarvisPrompts+LocalAgent.swift:9` still says *"The capture_screen, speak, and stay_silent names below are an output JSON protocol, not callable Codex tools."* That text ships live as Codex's base instructions (`CodexAppServerRuntime.swift:294`, `CodexExecRuntime.swift:200`).

`search_prep_notes` isn't in that list. So on a Codex session with prep material, Codex may try to run it as a real tool instead of just emitting the name in its JSON reply. It's the same stale-list problem this PR fixes in the coach prompt — better to build the list from the `tools` array.

### 3. Only mention the tool when it's actually offered

The `# Prep material` section sits in the system prompt on every session, including the default one where no prep material exists. That invites the model to call a tool it doesn't have — and that call is a hard attempt failure, three of which exhaust a route target. Describe only the tools the model actually has.

### 4. The model can't tell when it should search

The prompt says to search when a question *"resembles a topic in the user's prepared notes"* — but the model never sees the notes. No filenames, no topics. With `.required` tool choice, the safe move is to search on almost anything, so I'd expect this to fire on most attempts rather than the relevant ones.

That's expensive: every search adds a full model round-trip on the live path, which #30 already flagged as the main latency cost. Meanwhile the search itself is free — in-memory BM25 over a handful of files. We're spending the scarce thing to save the cheap thing.

Two options, neither blocking this PR:

- **Show a table of contents at session start.** Per source: the filename plus a short digest (headings, or the top terms the index already computes), ~50–100 tokens each. Then the model can skip the search when nothing matches, and write a better query when something does. This is how Claude's Agent Skills work — names and descriptions up front, full content only on demand.
- **Just run the search automatically** on each trigger using the new speech as the query, and include results in the first request when the score is high enough. No extra round-trip. Keep the tool for follow-up queries.

Worth weighing against: BM25 scores are hard to threshold, irrelevant chunks add noise, and the explicit tool call does give clean Activity provenance plus a better query than raw transcript text.

Refs: [Claude Code doesn't index your codebase](https://vadim.blog/claude-code-no-indexing/), [Agent Skills progressive disclosure](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview).

EdHasNoLife:
Thanks for the thorough review. Fixed 1–3:

**1. Observations clobbered** — fixed. `search_prep_notes` now appends to `work.observations` instead of replacing it, so a prior `capture_screen` result (tool-loop or prepared manual hint) survives a later search in the same attempt. Added a regression test covering capture → search → failure → retry, asserting both the image and the search result land in the retried request.

**2. Codex told the wrong tool list** — fixed. Rather than patch the enumeration to add `search_prep_notes` (which just defers the same staleness to the next tool), reworded `codexDirectResponse` generically: it now says *any* tool-shaped name in the per-turn protocol is an output-JSON label, never a real Codex tool — regardless of which names exist. Should hold without further edits as the tool set grows.

**3. Prep-material guidance always in the prompt** — fixed. Pulled the "# Prep material" section out of the static system prompt into a separate `prepMaterialAddendum`, appended only when `attempt.prepMaterial != nil`. A session with no prep material now gets a coach prompt that never mentions the tool at all — matches what its `tools` array actually offers. Added tests for both the presence and absence case.

**4. Relevance/latency tradeoff** — good catch, and I agree it's real: with `.required` tool choice and no visibility into what's actually in the notes, the model has little basis to skip the search, so it likely fires more often than it should, and each firing costs a full round-trip on the live path. Not fixing this in this PR per your own framing — it's a genuine design question (table-of-contents-at-Start vs. automatic-search-as-part-of-the-first-request are both reasonable, with real tradeoffs either way) that deserves its own scoping pass rather than a reactive patch here. Checking with the feature owner on whether to track it as a follow-up issue now or fold it into the #219 slice's own retrospective.

847/847 tests + both guards pass on `f2cf04f`.

EdHasNoLife:
Self-audit follow-up, prompted by "what else could we have missed": the "append instead of replace" fix for `search_prep_notes` (previous comment) only considered carry-forward *within* one attempt. It missed that a failed attempt's `work` carries forward **verbatim** into its retry (`work = failedWork` in `CoachDriver.swift`) — so a second `search_prep_notes` call on that retry would append its result on top of the first search's now-stale one instead of replacing it, slowly accumulating duplicate/outdated search results across a failure chain.

Replaced the flat `observations: [ChatMessage]` array with two independent slots — `screenObservation` and `prepNotesObservation` — each only ever touched by its own tool, so neither can clobber or pile up on top of the other, in either direction, across any number of attempts. Added a test with two searches across two failed attempts, asserting only the second's result survives into the final retry.

848/848 tests + both guards pass on `512dae6`.

JINGBANZ (approved):

Issue #231, opened 2026-08-29
fix(shortcuts): manual hint hotkey can silently drop a press during a live session


EdHasNoLife:
## Summary

The manual hint hotkey (⌥⌘J) can silently fail to fire during a live, active coaching session: registration succeeds, the session is running, the user presses the shortcut, and nothing happens — no hint, no beep, and no trace at any logging layer. Reproduced and confirmed against a real session's on-disk logs, not just reported behavior.

## Evidence

Session `2026-08-29_15-58-02_873F` (`~/Library/Application Support/Jarvis/sessions/2026-08-29_15-58-02_873F/`), a live ~6-minute coaching session with continuous turnEnd/silence trigger activity throughout:

- **Registration succeeded at launch**: neither of `HotkeyController`'s two failure-path log lines appears anywhere in the app's lifetime (checked via `log show` from process launch through the session) — `"hint-hotkey handler install failed"` (`HotkeyController.swift:50`) and `"hint hotkey ⌥⌘J unavailable"` (`HotkeyController.swift:62`) are both absent, meaning `InstallEventHandler`/`RegisterEventHotKey` both returned `noErr`.
- **The session was live**, not stopped — `AppDelegate.swift:200-206` plays `NSSound.beep()` as the explicit fallback when the hotkey fires with no `requestManualHint` closure set (i.e. session stopped); no beep occurred, and the log shows continuous coaching activity (turnEnd/silence triggers, brain calls) across the window the user pressed the key in.
- **No trace of the press reaching the coaching pipeline**: `CoachDriver` logs `"⌨️ hint shortcut — ..."` (referenced in #48) the moment a `.manualHint` trigger is actually handled — this line does not appear anywhere in the session's `jarvis-debug.log`, despite the user reporting a press mid-session.

So the press was lost somewhere between the Carbon key-event callback (`HotkeyController.swift:41-48`, which synchronously calls `onRequestHint?()` with no logging of its own) and `CoachDriver.handleTrigger(.manualHint)` actually running — with zero diagnostic trail at any point in between.

## Suspect area

[...]

PR #234, opened 2026-08-30
fix(shortcuts): trace and harden the manual hint hotkey against silent drops

EdHasNoLife:
## Summary

Investigates #231 — the manual hint hotkey (⌥⌘J) silently failing to fire mid-session, with zero trace at any logging layer.

I read through `HotkeyController`, the `AppDelegate` hotkey plumbing, `TurnTaskBox`, and `CoachDriver`'s trigger-coalescing/single-flight logic (`claimOrPend`, `selectBrainForAttempt`, `handleTrigger`) — the area #231 names as the suspect. That mechanism looks logically sound: every reachable branch of `claimOrPend` (`.claimed`, `.pending`, `.exhausted`) already `jlog`s, and a pended manual hint is correctly coalesced with priority and carried forward — a regression test below confirms it still executes once the in-flight attempt completes. I did **not** find a confirmed drop inside `CoachDriver`.

What I did find is a real gap: there was **no logging at all** between the Carbon hot-key callback firing and the press reaching (or not reaching) the coaching pipeline — exactly the blind spot #231's own step 1 called out. This PR closes that gap plus one theoretical (currently unreachable, but silent) path, and adds the regression test #231 asked for.

## Changes

- `HotkeyController`: log unconditionally the moment the Carbon callback fires — proves the OS-level event was delivered, before it depends on `onRequestHint` being wired up correctly.
- `AppDelegate`: log which of the two outcomes the press took — routed into the running session, or beeped because no session is live.
- `CoachDriver.selectBrainForAttempt`: log the one branch that could still return silently (a trigger claimed the instant before the route exhausts) with the same "route already exhausted" message used elsewhere, so a hotkey press can never fall through without a trace anywhere.
- `CoachDriverManualHintTests`: new regression test — a manual hint pressed while a turn-end attempt is in flight returns `.busy` immediately, then the *original* `handleTrigger` call goes on to run the forced hint (screenshot + forced `speak`) once the in-flight attempt completes.

## Why not more?

#231's completion criteria is satisfied: a press now always either executes, beeps, or is logged as pended — never silently disappears — and that's provable in the logs. Since I couldn't reproduce or definitively root-cause the original drop from static reading alone, the honest move was to add the missing instrumentation rather than guess at a fix for a mechanism that already looks correct. If it recurs, the new `HotkeyController`/`AppDelegate` logs will show whether the drop is before or after the app boundary.

## Test plan

- [x] `swift build && ./scripts/run-tests.sh` — 858 tests pass, including the new `manualHintPressedWhileAnAttemptIsInFlightStillExecutesAfterItCompletes`.
- [ ] Live smoke: press ⌥⌘J during a live session and confirm the two new log lines appear in `jarvis-debug.log`.

### DIFF
--- a/Sources/JarvisApp/App/AppDelegate.swift
+++ b/Sources/JarvisApp/App/AppDelegate.swift
@@ -199,9 +199,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate, BrainCompositionHost {
         hotkeys = HotkeyController()
         hotkeys?.onRequestHint = { [weak self] in
             guard let self, let fire = self.requestManualHint else {
+                jlog("Jarvis: hint hotkey ignored — no live session, beeping instead.")
                 NSSound.beep() // ghost-mode-allowed: explicit user hotkey while stopped
                 return
             }
+            jlog("Jarvis: hint hotkey routed into the coaching pipeline.")
             fire()
         }
 
--- a/Sources/JarvisApp/Shortcuts/HotkeyController.swift
+++ b/Sources/JarvisApp/Shortcuts/HotkeyController.swift
@@ -42,8 +42,12 @@ final class HotkeyController {
             guard let userData else { return noErr }
             let controller = Unmanaged<HotkeyController>.fromOpaque(userData).takeUnretainedValue()
             // Carbon delivers application-target hot-key events on the main thread, so it is safe to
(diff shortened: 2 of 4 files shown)


EdHasNoLife:
Closing without merging per request — verified working via code review, tests, and a live smoke test (see PR discussion), but not landing this change.

PR #242, opened 2026-08-31
feat(transcription): add vocabulary keyword hints for GPT Transcribe/Live

EdHasNoLife:
## Summary

- The reported transcription-accuracy failure mode was mangled technical terms and names (no
  code-switching involved), which is exactly what OpenAI's GA Realtime `keywords` field exists to
  fix.
- Adds a free-text, comma-separated vocabulary glossary to Settings -> Transcription
  (`TranscriptionControls`), persisted via `TranscriptionPreferences.openAIVocabularyKeywords`.
- Threaded through `TranscriptionConfiguration` -> `TranscriptionSessionFactory` ->
  `RealtimeTranscriber` -> `RealtimeSession.sessionUpdate`, sent as `keywords` (distinct from
  `prompt`) only for `gpt-transcribe`/`gpt-live-transcribe`.
- `gpt-4o-transcribe` (the default model) has no `keywords` or `prompt` field and is untouched; the
  Settings row says so when that model is selected. Neither the default model nor the default
  expected-language (Automatic) was changed -- both are already switchable from existing pickers.
- `wiki/decisions.md` (new dated entry plus supersession backlinks on the two prior entries that
  explicitly deferred vocabulary keywords), `wiki/architecture.md`, and `wiki/settings-window.md`
  updated per the wiki keep-in-sync checklist.

## Test plan

- [x] Gate (`swift build && ./scripts/run-tests.sh`) passes: 874 tests, 102 suites.
- [x] New/updated tests: `RealtimeSessionTests.keywordsOnlySendToCommittedTurnModels`,
  `TranscriptionPreferencesTests` (round-trip + blank-entry trimming), `DefaultsTests`.
- [ ] Not manually exercised in the running app -- the new Settings UI row (text field, show/hide,
  detail-text switching) was verified by build/tests only, not by clicking through Settings in a
  live session.

Closes #240. The deferred per-session/prep-material-derived vocabulary layer is tracked separately
in #241.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a Vocabulary glossary setting for transcription, supporting comma-separated keywords.
  * Glossary entries are cleaned and saved automatically when editing is completed.
  * Vocabulary hints are sent for GPT Transcribe and GPT Live Transcribe.
  * GPT-4o Transcribe does not use vocabulary hints.

* **Documentation**
  * Updated transcription settings and architecture guidance, including glossary behavior, model support, and screen-capture OCR details.

[...]

PR #244, opened 2026-09-01
test(diagnostics): fix flaky JarvisLog cross-suite attachment race

EdHasNoLife:
## Summary

- `DiagnosticEvidenceTests.aFullMailboxDropsOneDiagnosticAndKeepsAdmittingLaterOnes` flaked once in
  CI (`healthMarker(in:)["queue_overflow"]` came back wrong) and passed cleanly both locally and on
  an immediate CI rerun of the exact same commit — the signature of a race, not a logic bug.
- Root cause: `JarvisLog`'s attachment (`Sources/JarvisCore/Diagnostics/Log.swift`) is one
  process-global slot. Every suite that installs it for a test (`DiagnosticEvidenceTests`,
  `CoachDriverPipelineTests`, `CaptureHeartbeatTests`, `SessionAuditIsolationTests` via
  `CoachingParityHarness`) is already `@Suite(.serialized)`, but that only serializes cases *within*
  one suite. swift-testing still runs distinct suites concurrently by default, so a test in one suite
  could attach its own session in the narrow window between another suite's test attaching and
  reading its assertions back — silently repointing `jlog` traffic to the wrong session and
  corrupting its counters.
- Fix: `JarvisLogAttachmentLock` (`Tests/JarvisCoreTests/Diagnostics/JarvisLogAttachmentLock.swift`),
  a semaphore-backed acquire/release pair held around every `JarvisLog.attach`/`detach` span across
  all four affected files, so concurrent suites serialize around the shared slot instead of racing
  for it.
- **Test-only fix** — `Sources/JarvisCore/Diagnostics/Log.swift` (production `JarvisLog`) is
  unchanged. Making `attach` itself block until a prior session detaches was considered and rejected:
  production Stop does not guarantee a `detach()` call (a sealed handle already refuses late events),
  so a real second session's `attach()` at the next Start could deadlock waiting for a `detach()`
  that never comes.
- Logged in `wiki/decisions.md` (2026-09-01 entry) with the rejected alternatives (blocking
  production `attach`, disabling test parallelism repo-wide, merging the four suites into one).

## Test plan

- `swift build && ./scripts/run-tests.sh` — 872 tests, 102 suites, all passing (run 4x locally to
  check for regressions/new flakiness; all green).
- `./scripts/check-ghost-mode.sh` — passed (no production presentation code touched).
- Note: the original failure only reproduced once in CI and never locally, so this PR can't
  definitively prove the race is gone — only that the identified synchronization gap is closed and
  the fix doesn't regress anything. Worth watching CI on this branch and #242/#243 for a few more
  runs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic and coaching test reliability by preventing concurrent log attachments from interfering with one another.
  * Ensured log evidence is consistently captured and cleaned up, including when test operations encounter errors.
  * Reduced test execution delays caused by contention during shared log access.

* **Documentation**
  * Documented cross-suite log attachment synchronization behavior, limitations, and safe handling across asynchronous test operations.

### DIFF
--- a/Tests/JarvisCoreTests/Coach/CoachDriverPipelineTests.swift
+++ b/Tests/JarvisCoreTests/Coach/CoachDriverPipelineTests.swift
@@ -441,29 +441,29 @@ final class FakeOverlay: OverlayRendering, @unchecked Sendable {
         // what makes this boundary worth pinning: the two projections must still stay separate.
         let (activityLog, evidence) = ActivityLog.recordingSession(in: dir)
         defer { activityLog.disable(); try? FileManager.default.removeItem(at: dir) }
-        JarvisLogAttachmentLock.acquire()
-        defer { JarvisLogAttachmentLock.release() }
-        JarvisLog.attach(to: evidence)
-        defer { JarvisLog.detach() }
+        try await JarvisLogAttachmentLock.withExclusiveAttachment {
+            JarvisLog.attach(to: evidence)
+            defer { JarvisLog.detach() }
 
-        let brain = ScriptedBrain(script: [
-            .init(toolCalls: [.speak(callId: "s", lines: ["activity-boundary-tip-417"])])
-        ])
-        let (driver, _) = makeDriver(
-            activity: evidence, brain: brain, clock: ManualClock(now: 417))
+            let brain = ScriptedBrain(script: [
+                .init(toolCalls: [.speak(callId: "s", lines: ["activity-boundary-tip-417"])])
+            ])
+            let (driver, _) = makeDriver(
+                activity: evidence, brain: brain, clock: ManualClock(now: 417))
 
-        #expect(await driver.handleTrigger(.silence(secondsQuiet: 417)) == .spoke)
+            #expect(await driver.handleTrigger(.silence(secondsQuiet: 417)) == .spoke)
 
-        _ = await evidence.close()   // barrier: drains this session's accepted rows
-        let jsonl = try String(contentsOf: dir.appendingPathComponent("jarvis-activity.jsonl"), encoding: .utf8)
-        #expect(jsonl.contains("activity-boundary-tip-417"))
-        #expect(!jsonl.contains("quiet for 417s"))
-        #expect(!jsonl.contains("thinking"))
+            _ = await evidence.close()   // barrier: drains this session's accepted rows
+            let jsonl = try String(contentsOf: dir.appendingPathComponent("jarvis-activity.jsonl"), encoding: .utf8)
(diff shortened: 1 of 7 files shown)


EdHasNoLife:
The first version of this fix (semaphore-backed \`JarvisLogAttachmentLock\`) still failed on the
same test in CI. On reflection that was a real risk in the design, not just an incomplete fix:
every call site holds the lock across real \`await\` points (\`evidence.close()\`,
\`waitForDebugLine\`, \`CoachingParityHarness.run\`'s own async work), and \`DispatchSemaphore.wait()\`
blocks the actual OS thread it runs on. Swift Concurrency's cooperative thread pool doesn't grow to
compensate for a thread stuck like that the way GCD's would, so on a CI runner with few cores this
risked starving the pool rather than cleanly serializing the intended critical section — plausibly
making things worse, not just failing to fix the original race.

Pushed a follow-up commit replacing it with an \`actor\`-based lock (suspends instead of blocking,
releases on every path including a thrown error). Every call site now wraps its whole
attach...detach span in \`JarvisLogAttachmentLock.withExclusiveAttachment { ... }\` instead of manual
acquire/release. Gate passes locally, run 5x clean. Watching CI on this branch now.

EdHasNoLife:
Re CodeRabbit's review (5074319925, against d25b14f — before the actor-based redesign in ddad125):

1. **`CoachingParityHarness.run` acquire/release imbalance** — no longer applicable. That finding
   describes the manual `JarvisLogAttachmentLock.acquire()`/`.release()` pattern from the first
   (semaphore) attempt, which ddad125 removed entirely in favor of
   `JarvisLogAttachmentLock.withExclusiveAttachment { ... }` — a closure-based helper that acquires
   and releases on every path (including a thrown error) with no manual pairing to get wrong.

2. **Stale `EvaluationTranscript`/`Log.swift` references "around line 2185-2189"** — checked against
   current content: this PR's new entry at that location references
   `Sources/JarvisCore/Diagnostics/Log.swift`, which exists and is correct, and doesn't mention
   `EvaluationTranscript` at all. There *is* a genuinely stale `EvaluationTranscript.swift` path
   elsewhere in `wiki/decisions.md` (lines 310/675/1339 say `Sources/JarvisCore/Diagnostics/...`;
   the real file is `Sources/JarvisEvaluation/EvaluationTranscript.swift`) — but that's pre-existing
   drift this PR never touches, not something introduced here. Leaving it out of scope for this fix;
   happy to file it separately if useful.

Gate passes, 872 tests, run clean multiple times locally.

JINGBANZ (approved):

Issue #247, opened 2026-09-01
Improve system-design interview coaching alignment


EdHasNoLife:
## Problem

Jarvis coaches technical interviews well for coding/algorithm rounds, but for system-design rounds
the coaching tip sometimes doesn't align with what's currently being discussed — e.g. the user is
working through functional requirements or API design, and Jarvis's tip is about something
unrelated. Reported by the user (2026-09-01).

## Current state (grounded in code, as of this filing)

- The coaching system prompt (`Sources/JarvisCore/Prompts/JarvisPrompts+Coach.swift`) identifies
  Jarvis generically as a coach for "behavioral, system-design, and coding interviews" but contains
  **zero system-design vocabulary** — no mention of functional requirements, non-functional
  requirements, API design, data model, high-level architecture, or trade-offs anywhere.
- Every coaching turn is decided independently from turn-local signals (silence, stuck-ness, direct
  address) plus a free-text `search_prep_notes` query the model writes itself against a generic BM25
  index (`Sources/JarvisCore/PrepMaterial/PrepMaterialIndex.swift`) — nothing constrains the query to
  the specific sub-topic currently active, so it can retrieve an irrelevant chunk.
- History compaction (`Sources/JarvisCore/Prompts/JarvisPrompts+HistorySummary.swift`) asks the model
  to preserve "the active topic and load-bearing details" in a free-text briefing — unclear whether
  this reliably tracks the active sub-topic through a multi-phase system-design discussion, or
  whether it's actually a contributor to the reported drift.
- **Prior art**: `wiki/decisions.md` (2026-08-07, "History compaction is language- and
  interview-format neutral") explicitly *rejected* a hard-coded interview-phase detector/state
  machine — an automatic, per-turn, in-session classifier that guesses which phase the conversation
  is in — as "another brittle state machine" that risks misclassifying mixed-format sessions. Any
  proposal here needs to respect that rejection or make a clear case for revisiting it.

## Candidate directions (discussed, not yet decided)

1. **Add system-design domain vocabulary to the coaching prompt** — the standard flow (functional
   requirements → non-functional requirements → API design → data model → high-level architecture →
   deep dive/trade-offs) as reference knowledge, with an instruction to infer which part the
   candidate is currently addressing from the live transcript and keep the tip scoped there. Prompt
   content only — no new state, no per-turn classifier, does not conflict with the prior rejection
   (nothing external tracks or enforces phase; the model reasons in-context with vocabulary it
[...]

PR #252, opened 2026-09-02
feat(coach): add an interview-format picker with a system-design vocabulary addendum

EdHasNoLife:
## Summary

Phase 2 of #247 (system-design coaching alignment) — the main piece. Phase 1 (prep-notes query
scoping) is #251, still in review.

- The coaching system prompt had **zero** system-design vocabulary anywhere — no functional
  requirements, non-functional requirements, API design, data model, architecture, or trade-offs.
  Tips could drift from whatever stage of a system-design discussion the candidate was actually in.
- New `CoachingSkill` protocol + `InterviewFormat` enum (`coding` / `systemDesign` / `behavioral`),
  one enum with computed properties conforming to the protocol — matches how every other "one of a
  few known variants" setting in this codebase is modeled, not a struct per case. The protocol
  exists so a structurally different kind of skill could conform on its own type later without
  touching `InterviewFormat` — deliberately smaller than a runtime-pluggable skill system, which
  nothing today needs.
- Picked once in Settings → Brain's Coaching card (new "Interview format" row next to Reasoning
  Effort), applies on the next Start, never reclassified mid-session. **No auto-detection**: an
  automatic classifier was considered and explicitly rejected — it collides with either failure mode
  the 2026-08-07 decision already rejected for history compaction (misclassifies a session that
  shifts formats if it locks in once, or reintroduces a "brittle state machine" if it re-guesses). No
  selection means every format's non-empty addendum is included instead of guessing.
- Only `.systemDesign` has real content: the standard flow (functional requirements → non-functional
  requirements → API design → data model → architecture → deep dive/trade-offs) plus an instruction
  to infer which stage the candidate is addressing and stay scoped to it. `.coding`/`.behavioral`
  stay empty — the user reports those already work well, so no new prompt content was written for
  them without evidence it's needed.

## A real correctness constraint, not just precedent-following

The coaching system prompt is assembled in **two places**: per-turn in `CoachAttemptRunner` for
OpenAI-style providers, and once at Start in `BrainComposition`, baked into a persistent process for
local CLI providers (Claude Code/Codex). `CLIBrainClient` asserts its instructions never change after
construction — a value that could change mid-session would need to update both sites in lockstep or
hard-fail every CLI turn. Since the interview format is fixed for the whole session, the resolved
addendum is computed once and fed identically to both sites, including a later mid-session brain
hot-switch (`applyBrainPreferencesToRunningSession`, which doesn't go through the App's Start snapshot
— the value lives as session-scoped state on `BrainComposition` itself, mirroring `activeBrainTarget`).

## Also touches a documented Non-Goal

`wiki/architecture.md` §6 listed "Multiple coaching modes" as a v1 non-goal. Narrowed, not removed:
still no tiered sensitivity dial and no genuinely different gating/tools/proactivity per mode — an
optional format selection now supplies additional prompt *vocabulary* only. Logged in
`wiki/decisions.md` (2026-09-01) with the full reasoning and rejected alternatives.

## Test plan

- `swift build && ./scripts/run-tests.sh` — 882 tests (10 new), 104 suites, all passing.
- `./scripts/check-ghost-mode.sh` — passed.
- New tests: `InterviewFormatTests` (display names, addendum content/emptiness, automatic
  resolution), `CoachDriverInterviewFormatTests` (system prompt includes/omits the addendum per
  selection, mirroring `CoachDriverPrepMaterialTests`'s pattern), `BrainPreferencesTests`/
  `DefaultsTests` round-trip additions.
- Not covered by tests: the actual Settings UI row rendering/interaction, and whether the CLI-provider
  byte-identical-prompt invariant holds against a real local-agent process — both need manual/live
  verification, which wasn't done here.

## Summary by CodeRabbit

* **New Features**
  * Added Interview Format options for Coding, System Design, Behavioral, or None.
  * Selected formats are saved and applied to new coaching sessions.
  * Sessions now receive guidance specific to the selected format; choosing None adds no format guidance.
  * Added system-design interview guidance and improved availability of bundled coaching resources.

* **Documentation**
  * Updated settings and architecture documentation for Interview Format behavior.

* **Tests**
  * Added coverage for selection, persistence, prompt guidance, and clearing the selection.


EdHasNoLife:
Addressed CodeRabbit's finding (eb2f152): the system-design addendum said "the candidate" one
sentence, then switched to "me" the next — a real internal inconsistency in my own wording (not
just the ambiguity CodeRabbit flagged in isolation). Fixed to say "the candidate" consistently
throughout.

Gate passes, 887 tests.

EdHasNoLife:
Follow-up (22c449e): moved the system-design skill content out of a Swift string literal into a real
Markdown file (\`Sources/JarvisCore/Resources/Skills/system-design.md\`), served through JarvisCore's
own \`Bundle.module\` at runtime. Discussed in chat — skill content should read/edit like prose, not an
escaped multi-line Swift string. Still developer-only/rebuild-required; a true self-service system
(drop in your own skill file, no code change) was discussed and set aside as unneeded speculative
infrastructure for now.

This also required fixing \`scripts/build-app.sh\` and \`scripts/package-app.sh\`: JarvisCore now has
its own resource bundle, and both scripts only copied JarvisApp's — without the fix the skill would
resolve fine under \`swift test\` but silently go missing from the actual shipped app. Verified by
running \`./scripts/build-app.sh\` for real and confirming the \`.md\` file lands inside the signed
\`.app\`, not just inferred from the test suite.

Gate passes, 887 tests.

JINGBANZ:
Nice work, and the `Resources/Skills/*.md` mechanism is the right shape: prompt content that reads
and edits like prose instead of a Swift string literal, and it extends cleanly as more formats get
written.

One blocking issue: the default changes behavior for every existing user.

`Defaults.Brain` registers the key with no value, so `interviewFormat` is `nil` on every install and
the popup lands on "Automatic". `resolvedPromptAddendum(for: nil)` then concatenates every non-empty
addendum, which today is exactly `system-design.md`. So without anyone touching Settings, every
session's prompt now ends with "This is a system-design interview." That is an assertion rather than
neutral vocabulary, and it mislabels the coding and behavioral sessions this PR deliberately leaves
alone because they already work well. It also doesn't scale: once the other two `.md` files have
content, Automatic concatenates three contradictory "This is an X interview" paragraphs.

Two changes:

- Replace "Automatic" with **"None"**, meaning an empty addendum, so nothing changes for a user who
  doesn't opt in. `resolvedPromptAddendum(for:)` goes away, since resolution is then just
  `format?.promptAddendum ?? ""`.
- Build the menu from formats that actually have content:
  `InterviewFormat.allCases.filter { !$0.promptAddendum.isEmpty }`. Today that reads
  None / System Design. Dropping in `coding.md` later adds Coding with zero Swift changes, and no
  entry is ever byte-identical to None.

Keep all three enum cases. Worth a test asserting that with no selection the coaching system prompt
equals `JarvisPrompts.Coach.system` byte for byte.

Most of the "why not an automatic classifier" prose can go with Automatic.

Filed the two-prompt-assembly-sites cleanup separately as #253, out of scope here.

EdHasNoLife:
Good catch — thank you. Implemented exactly as specified (2ec8010):

- "Automatic" → **"None"**; `resolvedPromptAddendum(for:)` deleted, resolution is
  `format?.promptAddendum ?? ""`.
- The picker is built from `InterviewFormat.allCases.filter { !$0.promptAddendum.isEmpty }`, so it
  only ever offers None + System Design today, and a future `<format>.md` needs zero Swift changes
  to appear.
- Added the byte-for-byte regression test: with no selection, the system prompt sent to the model
  equals `JarvisPrompts.Coach.system` exactly.
- Trimmed the wiki (architecture.md, settings-window.md) to describe the corrected default and
  explain why the union-of-non-empty design was wrong, not just different.

Saw #253 — agreed it's the right separate follow-up, not folding it in here.

Gate passes, 886 tests (net -1: replaced two tests covering the deleted function with one covering
the deletion contract, and swapped the old "None guesses system design" test for the byte-for-byte
one you asked for).

EdHasNoLife:
Follow-up (8404fe6), addressing CodeRabbit's two newer findings:

**"Move resource loading out of the coaching kernel" (13:13 UTC)** — real, and worse than the literal suggestion: `Bundle.module`'s generated accessor only checks `Bundle.main.bundleURL` (the `.app` root, not `Contents/Resources`, where the packaging scripts actually copy resources) or a compile-time absolute build path, and `fatalError`s if neither exists. A real install has neither — so this would have crashed the process the first time anyone selected a format with content. Fixed by replacing `Bundle.module` with a manual `skillMarkdownURL(named:)` search mirroring the existing `SileroVoiceActivityDetector.bundledModelURL()` pattern, plus a `#filePath`-relative fallback for `swift test` (whose `Bundle.main` resolves to the unrelated `swiftpm-testing-helper` toolchain binary, not this checkout — verified with a throwaway diagnostic test). That still leaves file I/O in `Sources/JarvisCore/Coach`, which `scripts/check-coaching-kernel.sh` forbids, so `InterviewFormat`, `CoachingSkill`, and their tests moved to `Config/` — a path the kernel check already excludes, and semantically correct: `CoachDriver`/`CoachAttemptRunner` only ever see a pre-resolved `String`, never the `InterviewFormat` type, so this is a pure relocation with no coupling.

**"Align this paragraph with the current picker contract" (wiki/architecture.md, 15:00 UTC)** — also real: the paragraph listed Coding/Behavioral as picker options despite their empty addenda being filtered out, and overclaimed that dropping in a new `<format>.md` alone creates a picker entry. Rewritten to match `settings-window.md`'s wording (None, plus one entry per format that actually has content) and to state the actual contract: a resource-only change fills in an *existing* case, but a genuinely new format needs a new `InterviewFormat` case and display name before any Markdown file surfaces it.

Gate passes: 886 tests, ghost-mode guard, coaching-kernel guard all green.

EdHasNoLife:
Follow-up (12d6eb5), addressing claude-review's three newer findings (15:19 UTC):

**"CoachingSkill speculative abstraction"** — real. Confirmed zero `any CoachingSkill` usage anywhere in `Sources/`/`Tests/`; every call site (`BrainSection.availableFormats`, `CoachAttemptRunner`, `BrainComposition`) already used `InterviewFormat` directly, and `displayName`/`promptAddendum` are already concrete members. Deleted `CoachingSkill.swift` and dropped the conformance — no behavior change.

**"wiki/status.md not updated"** — real. The interview-format picker was built but never recorded there. Added it to the `Sources/JarvisCore/Config/` bullet with a pointer to `InterviewFormat.swift` and `Resources/Skills/`.

**"wiki/architecture.md pastes code instead of referencing it"** — real, and my own regression from the previous commit (8404fe6), where I'd literally pasted the two Swift filter/fallback expressions while fixing an unrelated CodeRabbit finding. Replaced both with symbol references (`BrainSection.availableFormats`, `CoachAttemptRunner`'s system-prompt assembly) per wiki/AGENTS.md Convention 4. While in there I also caught and fixed a second staleness bug in the same paragraph: it still described the removed `Bundle.module` path instead of the manual `skillMarkdownURL(named:)` search from 8404fe6.

(The two 13:15 UTC findings — the "me" wording and the earlier framing of the CoachingSkill issue — were already superseded by 2ec8010/8404fe6 before this push; not re-actioned here.)

Gate passes: 886 tests, ghost-mode guard, coaching-kernel guard all green.

EdHasNoLife:
Follow-up (deb410a), addressing claude-review's three newest findings (15:36 UTC):

**"showsSeparator: false is on the wrong row"** — real, verified against `SettingsRowView.layout()`: the separator is drawn at `y: bounds.height - 1` in unflipped coords, i.e. the row's own **top** edge. `effortRow` sits above `formatRow`, so the flag needed to be on `effortRow` (to suppress doubling with the card header's own divider) and absent from `formatRow` (to show the inter-row divider) — exactly backwards from what was there. Fixed to match the established single-row-card pattern used everywhere else in Settings (e.g. `HotkeySection`).

**"wiki is copying implementation details that live in the code"** — real. The paragraph re-described `skillMarkdownURL(named:)`'s three filesystem candidates and the `Bundle.module` fatalError, all of which are already documented word-for-word in that function's doc comment. Trimmed to a pointer per wiki/AGENTS.md Convention 4.

**"speculative enum cases (.coding/.behavioral)"** — not actioned. It's accurate that they're unreachable through the UI today (filtered out of the picker since their addenda are empty), but this is a domain-modeling judgment call, not a bug: unlike the `CoachingSkill` protocol removed earlier (a genuine zero-consumer indirection layer), these are two enum cases for the three interview types this feature was scoped around from the start, already carrying inline rationale for why they're intentionally empty rather than accidentally incomplete. Reversing that is a scope call I'll leave to a human reviewer rather than take unilaterally.

Gate passes: 886 tests, ghost-mode guard, coaching-kernel guard all green.

JINGBANZ (approved):

PR #267, opened 2026-09-05
feat(coach): show private system-design diagram hints

EdHasNoLife:
## Summary

System Design coaching can now show a private architecture sketch alongside a high-level-design hint. The existing `speak` action carries optional Mermaid boxes and arrows, including for the manual-hint shortcut; other interview formats keep their text-only schema and output.

The sketch is a native image attachment in the existing scrollable overlay box. Graphs preserve their layout and scale proportionally to the available width and 60% of the overlay height, updating during resize. Settings → Overlay → Overlay Box has a persisted **Show diagrams** toggle, enabled by default; turning it off hides graph attachments while retaining text and history. Unsupported diagrams fall back to the text hint. This supports a bounded Mermaid flowchart subset, not the full Mermaid language or a continuously synchronized architecture canvas.

## Security and privacy impact

Diagrams use the same nonactivating, capture-excluded panel and its existing visibility/session lifecycle. Rendering is native and memory-only: no JavaScript, web renderer, network assets, additional window, or archived image. Diagram source follows existing brain-history/wire-audit handling; Activity retains the text hint. The selected interview format is frozen at Start, and diagram fields are ignored outside System Design. Capture, credentials, provider routing, and retry behavior are unchanged.

## Validation

- [x] `swift build && ./scripts/run-tests.sh` passes
- [x] swift-testing covers parsing/limits, format gating, forced manual hints, text fallback, capture-excluded attachments, proportional fit in both dimensions, immediate panel resizing, toggle persistence/hide/restore, and bypass routing
- [x] Conventional Commit
- Opened a signed synthetic demo with the actual overlay and Settings components at default window size and 50% opacity; inspected the Show diagrams row. Bypass routing remains covered by its regression test.

```text
Build complete! (8.00s)
Ghost-mode presentation API guard passed.
Coaching-kernel dependency guard passed.
Audio capture configuration guard passed.
App identity guard passed.
Release configuration guard passed.
Test run with 910 tests in 109 suites passed after 15.315 seconds.
```

### Live smoke

Live capture and real screen-sharing smoke were not run. The App change passes the Start-time format snapshot into the coach; a maintainer should check a System Design high-level-design hint and capture exclusion in the signed app before merge. Automated Overlay tests exercise the native panel and rendering using synthetic data.

## Summary by CodeRabbit

* **New Features**
  * System Design coaching can display private architecture diagrams alongside spoken tips.
  * Diagrams render as clear boxes and arrows with labeled connections and proportional scaling.
  * Added a **Show diagrams** setting for the Overlay Box, enabled by default and saved across launches.
  * Selected interview formats are preserved throughout session startup and coaching.
* **Bug Fixes**
  * Invalid or unsupported diagrams safely fall back to text-only guidance.
* **Documentation**
  * Added guidance covering diagram formats, limitations, privacy, and text-only behavior for other interview types.


JINGBANZ:
Two non-blocking followups, either here or as separate issues.

**1. Cache rendered diagrams (`Sources/JarvisOverlay/OverlayBoxPanel.swift:195`)**

`setEntriesText` rebuilds the whole attributed string and re-runs
`DiagramHintImage.render` for every entry carrying a diagram, including on plain
`append`. So a session with six diagram hints redraws all six at an unchanged width
on every subsequent tip, synchronously on the main actor, in the hint delivery path.
That waste grows linearly with session length.

Keying a cache on `(diagram, width)` makes the append path free and still
invalidates on resize, where a redraw is actually wanted.

**2. Move the mermaid contract into the tool description (`Sources/JarvisCore/Coach/ToolDefs.swift:37`)**

`systemDesignSpeakTool` declares `"required":["lines","mermaid"]` but reuses
`ToolDescription.speak` verbatim at :39, which never mentions the field. The only
thing describing a required parameter lives in `system-design.md`, so someone
rewriting the skill file gets no signal that a schema depends on it. This ToolDef
already carries its own description: move the mechanical rules there (flowchart
LR/TD, one statement per line, 12 boxes / 24 arrows, null for text-only tips) and
leave the judgment call in the skill file.

Neither blocks merge.

JINGBANZ (dismissed):

JINGBANZ:
Two things on 745c4e86. Neither blocks merge.

**1. Replace the flat 60% height budget with the measured text height (`OverlayBoxPanel.swift:213`)**

Fit-to-box is the right design, but `height: box.bounds.height * 0.6` is a guess where an
exact number is available. The reserve exists so the hint text and the sketch both fit
without scrolling, and the panel already knows how tall that text is: it is building the
attributed string. Measure the stamp and text run at the current width, subtract that plus
the insets, and pass the remainder.

The flat 60% is wrong in both directions. A one-line hint at the default 25pt font occupies
about 30pt, so in a 440pt box it reserves 176pt for 62pt of content. A hint that wraps to
four lines is about 120pt and may get clipped instead.

For a 4-box TD chain in a default 520x440 box:

| | available height | scale | label |
|---|---|---|---|
| shipped, flat 60% | 264pt | 0.36 | 5.5pt |
| measured reserve | ~378pt | 0.52 | 7.8pt |
| measured reserve + gap 50 | ~378pt | 0.66 | 9.9pt |

Which is the second half: at `boxSize.width = 172` and `gap = 100`, 37% of the natural
width is empty space, and under fit-to-box that whitespace is charged straight to the zoom
level. Halving the gap costs nothing visually.

Together that is 5.5pt to 9.9pt on the worst common shape, with the fit-to-box behavior and
the LR/TD orientation unchanged.

It does not rescue a six-box LR chain, and no scale factor does at that width. That case
belongs in the skill prompt, which currently says prefer 3 to 8 boxes.

**2. Two bits of redundant redraw, on different paths**

Moving `refreshText()` from `onEndLiveResize` to `onResize` means the message history is now
rebuilt on every frame of a drag rather than once on release. I measured that on this branch
rather than guessing:

| session contents | per resize frame | of a 16.67ms frame |
|---|---|---|
| 0 diagrams, 6 text tips | 1.568 ms | 9.4% |
| 1 diagram, 5 text tips | 2.005 ms | 12.0% |
| 6 diagrams, 6 text tips | 5.515 ms | 33.1% |
| 12 diagrams, 6 text tips | 8.067 ms | 48.4% |

Nothing exceeds the frame budget, so nothing stutters. The measurement does separate three
cases, though, and only one of them is legitimate:

- *Resizing with diagrams present*, about 0.6ms per diagram per frame. Legitimate. Under
  fit-to-box the sketch genuinely has to be re-scaled as the box changes. Leave it.
- *Resizing with no diagrams*, 1.568ms per frame. Waste. The rebuild exists only to re-scale
  diagram images; with none present the text view already reflows its own text through
  `widthTracksTextView`. Every user who never selects System Design pays this on every drag.
  Gate the rebuild on `entries.contains { $0.diagram != nil }`.
- *A new tip arriving.* `setEntriesText` re-renders every older diagram at a width that has
  not changed, so six diagrams in the session means six redundant redraws for a tip that
  needed none. This predates 745c4e86. Cache the rendered image on `(diagram, width)`.

Worth noting the cache does not help the drag case, since the width really does change each
frame. The two fixes are for different paths.

JINGBANZ (dismissed):

JINGBANZ (approved):

PR #268, opened 2026-09-06
feat(transcription): add Gemini as a third transcription provider

ironman-mark-I:
Adds Google Gemini (`gemini-3.5-transcribe-live`) as a third transcription provider alongside OpenAI Realtime and Apple Speech.

**Spec:** `docs/superpowers/specs/2026-09-04-gemini-transcription-design.md`
**Plan:** `docs/superpowers/plans/2026-09-04-gemini-transcription.md`

## What's here

Three refactors first, because two shared assumptions blocked a second key-based provider:

- **`refactor(config)`** — the secret store was OpenAI-only end to end (hardcoded filename, env var, and a single shared instance). It's now keyed by `Credential`, promoted from the existing readiness enum rather than duplicated. The OpenAI file path and env var are unchanged.
- **`refactor(transcription)`** — `requiresOpenAIAPIKey` couldn't express "Gemini ears + OpenAI brain", which needs two distinct keys. Replaced by `requiredCredentials(for:) -> Set<Credential>`, feeding the readiness gate directly.
- **`refactor(capture)`** — `TranscriptionAudioFormat.pcm16Mono` was named as a neutral contract but encoded OpenAI Realtime's 24 kHz requirement. The rate is now provider-derived, so a Gemini session downsamples 48 → 16 kHz in one exact 3:1 decimation instead of chaining a second stage. OpenAI and Apple Speech still receive 24 kHz with byte-identical timing.

Then Gemini itself: the persisted model/mode/language/vocabulary preferences, a pure unit-tested wire contract (`GeminiLiveSession`), the WebSocket adapter (`GeminiLiveTranscriber`), and the Settings UI (a Gemini key card in Connections, per-provider rows in Transcription).

Gemini's server finalizes turns itself, so the adapter carries **no ledger, no turn coordinator, and no client-commit path** — it reuses the already provider-neutral `TranscriptionCoachingCoordinator` and `RealtimeContinuityReporter`.

## Two bugs the tests couldn't catch

- **Binary frames.** Gemini's `BidiGenerateContent` sends JSON as *binary* WebSocket frames. The adapter matched only `.string`, so every server message — including the `setupComplete` handshake — was silently dropped and readiness timed out forever. Found by live smoke test, not by 932 passing tests: the unit tests stop at the pure-parser boundary and never cross the `URLSessionWebSocketTask` frame-type seam.
- **Rejected keys.** A bad key does *not* produce an HTTP 401 at upgrade, nor an `{"error": ...}` frame. The handshake succeeds and the server closes with **code 1008**. That was being treated as transient, so a user waited ~61s of backoff and was told "connection was lost". Now mapped to a terminal `.authenticationFailed` — the one case AGENTS.md permits exhausting a target immediately, being a proven permanent provider-boundary failure.

## Security

The Gemini key travels in the WebSocket URL query string, so `connectURL` is the only place it appears and `redactedEndpoint` is the only thing diagnostics print. Every transport failure path constructs a fixed self-authored reason rather than interpolating the caught error — `URLError` descriptions can embed the failing URL. Verified with `grep -rc "key=" .jarvis/*/jarvis-debug.log` → 0 files, across both a failing run (4 failed sockets, 3 reconnects) and a passing one.

## Verification

- `swift build && ./scripts/run-tests.sh` → **936 tests / 109 suites green**
- **Live smoke:** both sockets ready on socket #1 in ~310 ms, zero reconnects, utterances transcribed on both mic and system-audio sides, 134 interim frames correctly discarded
- Invalid key fails fast with the rejected-key reason; OpenAI transcription re-verified unregressed
- Whole-branch review clean; wiki updated (`settings-window`, `architecture`, `status`, `transcription-benchmark`)

## Known follow-ups (not in scope here)

- ~130 lines of socket-lifecycle helpers are near-verbatim between the two WebSocket adapters. The spec deferred extraction until a third streaming provider; with two working implementations now visible, this is better done deliberately in its own PR. Cross-reference comments were added to both files meanwhile, since the real hazard is silent divergence. That PR should also fold in the `readyTimer`/`pingTimer`/`pongTimer` main-queue-vs-lock inconsistency (inherited from `RealtimeTranscriber`) and a missing `stopped` re-check in `sendHealthPing`.
- Gemini has no benchmark arms — `Arm.model` is typed `OpenAITranscriptionModel?`. Documented explicitly in code and `wiki/transcription-benchmark.md` rather than left silently unsupported.
- Gemini sends `voiceActivity` frames the adapter doesn't consume. Wiring them into `updateTranscriptionWork` would suppress a coaching turn mid-utterance more reliably than the current local-audio inference.

https://claude.ai/code/session_01Diw4AKtt8yvZ5u4zQBSiEX

## Summary by CodeRabbit

- **New Features**
  - Added Gemini Live as a transcription provider with live WebSocket transcription and reconnect handling.
  - Added separate OpenAI and Gemini API-key management.
  - Added Gemini settings for model, languages, vocabulary, and verbatim or smart transcription modes.
  - Added provider-specific audio processing, including Gemini’s 16 kHz format.

- **Bug Fixes**
  - Improved credential validation and routing for selected providers.
  - Improved transcript filtering and transcription stability during buffering and reconnections.

- **Documentation**
  - Updated architecture and settings documentation with Gemini support and provider-specific requirements.


Sources/JarvisCore/Config/Secrets.swift
```diff
@@ -1,81 +1,85 @@
...
             // owner-only guarantee holds regardless. Best-effort: a metadata-perms failure must not
             // block saving the key, whose own bytes are still protected 0600.
-            try? fm.setAttributes([.posixPermissions: 0o700], ofItemAtPath: dir.path)
+            try? fm.setAttributes([.posixPermissions: 0o700], ofItemAtPath: directoryURL.path)
         } catch { return false }
         // Create the file 0600 from the start (createFile applies attributes atomically on creation),
         // so the secret is never briefly world-readable between write and chmod.
-        return fm.createFile(atPath: fileURL.path, contents: Data(key.utf8),
+        return fm.createFile(atPath: fileURL(for: credential).path, contents: Data(key.utf8),
                              attributes: [.posixPermissions: 0o600])
```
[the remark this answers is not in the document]
ironman-mark-I: Verified this empirically and it does not reproduce — declining the fix, with evidence.

The premise ("`attributes` applies only when it creates a file") is true, but the conclusion doesn't follow: on Apple Foundation, `createFile(atPath:contents:attributes:)` **unlinks and recreates** rather than truncating in place, so an "overwrite" always takes the create path and the attributes are applied.

Probe: create the file 0600, tamper the mode, then save again exactly as `FileSecretStore.setApiKey` does — tracking the inode:

| pre-existing mode | after overwrite | content | inode changed |
|---|---|---|---|
| 644 | **600** | updated | yes |
| 666 | **600** | updated | yes |
| 777 | **600** | updated | yes |

`inodeChanged=true` is the mechanism: it's a new file each time, so the 0600 attributes are never skipped.

Two further points:

1. `Tests/JarvisCoreTests/Config/FileSecretStoreTests.swift:81` (`overwriteReassertsOwnerOnlyPermissions`) already asserts precisely this scenario — chmod 0644, re-save, assert 0600 — and passes. It predates this PR, so the guarantee isn't new.
2. This PR doesn't change that call. The same `createFile` invocation existed before; only the filename is now credential-derived. So even if the concern held, it wouldn't be a regression introduced here.

Also worth noting the quoted web result hedges: *"does not explicitly merge or preserve existing attributes **unless specifically handled by the underlying implementation**"* — and the underlying implementation does handle it.

Not applying the suggested change: a pre-write `chmod` would be unreachable code on the security-critical path, and treating its failure as a save failure would add a new way for saving a key to fail for a condition that cannot arise.

Sources/JarvisApp/Capture/GeminiLiveTranscriber.swift
```diff
@@ -0,0 +1,708 @@
...
+            self.updateWorkFlag()
+            self.pumpIfPossible()
+        }
+    }
+
+    /// `true` while buffered audio has not yet been accepted by a live socket — the coordinator holds
+    /// a turn open until this settles, since Gemini's transcript for that audio is still in flight.
+    private func updateWorkFlag() {
+        lock.lock(); let hasPendingWork = !bufferedAudio.isEmpty; lock.unlock()
+        coachingCoordinator.updateTranscriptionWork(hasPendingWork)
```
EdHasNoLife: [P1] Keep transcription unsettled until Gemini finishes recognition

This flag tracks only unsent PCM. Successful sends empty that queue before Gemini finishes recognition, so a healthy socket reports settled while an utterance is still in flight. If the other speaker's final arrives first, its batching timer can admit an automatic coaching attempt without the pending utterance: AppDelegate forwards this flag directly into CoachDriver's cross-speaker settlement gate. Pending work must include unfinished speech/recognition rather than clearing solely on send completion. This breaks the existing speech-gating contract even without a transport failure.

Sources/JarvisApp/Capture/GeminiLiveTranscriber.swift
```diff
@@ -0,0 +1,708 @@
...
+            // is routine, so byte-identical chunks are common, and equality could evict a same-bytes
+            // chunk that replaced the one actually sent while `lock` was released for `task.send`. If
+            // the token no longer matches the head, that chunk was evicted — do nothing to the queue.
+            if isCurrentLease, let first = self.bufferedAudio.first, first.token == token {
+                self.bufferedByteCount -= self.bufferedAudio.removeFirst().data.count
+            }
+            self.isSending = false
+            self.lock.unlock()
+            self.updateWorkFlag()
+            self.pumpIfPossible()
```
EdHasNoLife: [P2] Reject stale send completions before mutating send state

Queue removal checks isCurrentLease, but clearing isSending and pumping do not; the error branch at line 337 also clears isSending unconditionally. If a delayed completion from a replaced socket arrives while its replacement has a send in flight, it clears the replacement's flag. The following pump (or next capture callback on the error path) can then send the replacement's still-queued head again, duplicating PCM and allowing overlapping sends. Return on a stale task/generation before any send-state mutation or pumping, on both success and failure paths.

EdHasNoLife (dismissed):
Two correctness issues remain in the Gemini adapter: transcription settlement is cleared before recognition finishes, and stale send callbacks can mutate replacement-socket state.

Validation: `swift build && ./scripts/run-tests.sh` passed with 937 tests in 109 suites (exit 0). Live API/audio smoke testing was not performed.

Sources/JarvisApp/Capture/GeminiLiveTranscriber.swift
ironman-mark-I: Confirmed and fixed in `73dd504` — thank you, this was the most valuable finding on the PR.

You were right that the root cause is structural, not a slip. The spec for this branch asserted that Gemini "needs no ledger" because its server owns turn ordering. That's true for ordering, but the OpenAI ledger was also carrying the pending-*recognition* signal (`ledger.hasPendingItems` in `RealtimeTranscriptionLifecycle.swift:441`), and dropping it silently dropped half the settlement contract. Four earlier review passes missed it because they each checked the code against that spec.

The fix adds a `recognitionInFlight` condition alongside the existing unsent-PCM one, driven by Gemini's interim frames:

- an `interimInputTranscription` frame means Gemini is actively recognizing → in flight
- a finalized `inputTranscription` frame clears it
- silence emits no interim frames, so silence still settles correctly

Only the boolean is read; the interim text is never touched, so it cannot reach Activity or the transcript.

Two things I deliberately did *not* do, in case they look like omissions:

- **Not** "audio sent since the last final" — silence never produces a final, so that flag would wedge true and automatic coaching would never fire again.
- **Not** `voiceActivity` frames, which Gemini does send and which would tighten the leading edge (speech starts slightly before the first interim). Their payload shape isn't verified against the live API yet, so that's noted as a follow-up rather than guessed at here.

The flag is bounded — cleared on socket teardown, on reconnect, and in `stop()` — so a final lost to a dropped socket can't block coaching for the rest of the session, which would have been worse than the original bug.

Sources/JarvisApp/Capture/GeminiLiveTranscriber.swift
ironman-mark-I: Confirmed and fixed in `73dd504`.

Both paths now return early on a stale task/generation *before* any send-state mutation, rather than gating each mutation separately — so nothing can partially apply:

- **success path**: `guard isCurrentLease else { unlock; return }` ahead of the queue removal, the `isSending` clear, and the pump.
- **error path**: `isSending` is cleared only when the lease is current.
- **encode-failure path**: same guard added. It runs synchronously before the send so it was less exposed, but it shouldn't be the one place with a weaker rule.

One deliberate difference from your suggested shape, on the error path: `failConnection` is still called unconditionally rather than being skipped by the early return. It carries its own generation guard, so for a stale lease it's already a no-op — and that guard, not an early return here, is what keeps a stale diagnostic from disturbing a healthy replacement socket. Keeping the call preserves the transport-failure reporting you get today. There's a comment at the site recording that reasoning so it doesn't get "tidied" into an early return later.

Sources/JarvisApp/Capture/GeminiLiveTranscriber.swift
ironman-mark-I: Follow-up on this in `a4acff2` — I probed the live endpoint to check whether `voiceActivity` frames were a better signal than interims, and they partly are, so the fix now uses both.

Captured frame sequence with synthetic speech (identical in VERBATIM and SMART):

```
setupComplete
voiceActivity {"type":"ACTIVITY_START","audioOffset":"0.280s"}   <- top-level, sibling of serverContent
serverContent.interimInputTranscription {"text":"Let"} ...        <- ~500ms AFTER activity start
[all audio sent]
serverContent.inputTranscription  (FINAL)                         <- ~470ms after the last audio was sent
serverContent.generationComplete
voiceActivity {"type":"ACTIVITY_END",...}                         <- ~13ms AFTER the final
```

Three things fell out of it:

1. That ~470 ms gap between "last audio sent" and the final quantifies the bug you found — the pre-fix code settled 467 ms before recognition actually finished.
2. `ACTIVITY_START` beats the first interim by ~500 ms, so interim-only still left a leading-edge hole where the flag read settled while speech was already being recognized. `recognitionInFlight` is now set by **either** signal.
3. SMART does **not** suppress interims, which retires a risk I had been carrying — a mode that stopped emitting them would have silently disabled the interim-based gate.

`ACTIVITY_END` is deliberately **not** wired to anything. It landed 13 ms after the final in this capture, but that ordering is coincidence rather than a documented contract, and clearing on it would reintroduce the same settle-before-recognition race. The finalized transcript remains the only clear signal, with the existing bounding (teardown / reconnect / stop) unchanged so a lost final cannot wedge the flag. There is a comment at the field recording this, since "completing the pair" looks like tidying and would be a silent regression.

944 tests / 109 suites green; kernel guard passing.

Sources/JarvisApp/Capture/GeminiLiveTranscriber.swift
```diff
@@ -0,0 +1,768 @@
...
+            reportTerminalFailureOnce(failure)
+            return
+        }
+        if GeminiLiveSession.isSetupComplete(message) {
+            markReady(task: task, generation: socketGeneration)
+            return
+        }
+        if let text = GeminiLiveSession.finalTranscript(from: message, speaker: speaker) {
+            // This utterance is done — recognition is no longer in flight for it.
+            lock.lock(); recognitionInFlight = false; lock.unlock()
```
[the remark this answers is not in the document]
ironman-mark-I: Confirmed and fixed in `dbac63e`. Good catch — and worth noting this was a regression introduced by the settlement fix two commits earlier, not pre-existing: before `recognitionInFlight` existed there was no flag to wedge.

Root cause is that `finalTranscript(from:speaker:)` answers two different questions with one `nil`:

1. this frame is not a finalized `inputTranscription` at all, or
2. it *is* one, but `TranscriptFiltering` rejected its text.

Clearing the flag inside `if let text = finalTranscript(...)` therefore skipped case 2. And that case is the likely one, not an edge: `TranscriptFiltering.hallucinationDenylist` exists precisely because these models emit `"Thank you."` and punctuation-only finals **on silence**, so the exact scenario the filter was written for is the one that wedged the flag.

The fix adds a pure `hasFinalizedTranscription(_:)` presence predicate to `GeminiLiveSession`, symmetric with the existing `hasInterimTranscription(_:)`, and splits the two concerns:

- **raw finalized frame** → clear `recognitionInFlight`, then `updateWorkFlag()`. Recognition is over the moment any finalized frame arrives, whatever we decide to do with the text.
- **accepted text only** → `recordFinalizedTranscript`, the continuity-reporter server-speech event, and the benchmark observation.

So filtering still does its real job — keeping hallucinated text out of Activity and the transcript — without distorting the settlement signal. `ACTIVITY_END` continues to clear nothing; the finalized frame and the three bounding sites (teardown / reconnect / stop) remain the only clears.

The regression test asserts both halves at once, since that pairing is what the bug collapsed:

```swift
#expect(GeminiLiveSession.hasFinalizedTranscription(thankYou))            // still a final
#expect(GeminiLiveSession.finalTranscript(from: thankYou, speaker: .me) == nil)  // text still rejected
```

948 tests / 109 suites green; kernel guard passing.

ironman-mark-I:
Confirmed and fixed in `0f104ea`. `handle` now opens with a full lease revalidation instead of a `stopped`-only test:

```swift
guard isCurrent(task: task, generation: socketGeneration) else { return }
```

`isCurrent` is `!stopped && !isReconnecting && currentTask === candidate && generation == candidateGeneration`, so it strictly subsumes the check it replaces. `receiveLoop` keeps its own pre-decode check — one rejects early, the other rejects after the parse gap.

Agreed on the severity ranking, and the terminal-error path is the one that made this worth fixing rather than tidying: `reportTerminalFailureOnce` takes no generation and guards only `!stopped, !terminalFailureReported`, so it cannot self-guard the way `markReady` does. A dead socket's error frame would latch `terminalFailureReported`, emit `.failed`, and tear down the healthy replacement session. The `recognitionInFlight` and stale-transcript cases are real but recoverable by comparison.

`markReady`'s own `isCurrent` check is now redundant; I left it as defence in depth rather than removing it.

Two notes:

**No test.** This race lives inside the `URLSessionWebSocketDelegate` layer and there is no socket-level fixture for this class, so there is nothing to assert against without building a fake-transport seam — out of scope here, and I would rather say so than add a test that passes vacuously. It joins the known coverage gaps for this adapter (the `sendFrame` stale-completion race, reconnect exhaustion, pong timeout, buffer eviction at 16 kHz).

**`RealtimeTranscriber` has a narrower version of the same shape**, and I deliberately did not touch it in this PR. Its `receiveLoop` also checks `isCurrent` and its `handle` also tests only `stopped` — but its parse happens *inside* `handle`, after that test, rather than between the two checks, so the window is much smaller. Whether each of its downstream handlers re-validates after the parse is a separate audit, and changing the production OpenAI path inside a Gemini PR seemed like the wrong trade. Flagging it rather than silently fixing or silently ignoring it.

948 tests / 109 suites green; kernel guard passing.

JINGBANZ:
Reviewed at 0f104ea from a few angles. The settlement and lease fixes look
right, so this is only what's still open. Checked against the code at head and
Google's current docs; nothing here overlaps an existing thread.

Blocking in my view: 1 and 4, plus 5 depending on the answer. The rest are
follow-ups.

**1. The connection dies every ~10 minutes and we don't handle it.**
Google caps a Live connection at around 10 minutes and sends a `goAway` message
with `timeLeft` first. Neither `goAway` nor session resumption appears in the
adapter. Reconnect covers it so nothing breaks, but both sockets open together
and therefore cut together, and the utterance in flight loses its final
transcript while its audio replays from the middle. That's a garbled line every
10 minutes on both sides. Four times in a 45 minute interview, which is the
actual use case. Either handle `goAway` with a graceful rotation, or make it an
explicit documented limitation of the Gemini option. It shouldn't ship silently.
https://ai.google.dev/gemini-api/docs/live-api/session-management

**2. Close code 1008 isn't only a bad key.**
`GeminiLiveSession.swift:171` maps every 1008 to `.authenticationFailed`. 1008
is Google's generic rejection code and also carries "model not found for this
API version" and "not supported". The `reason` payload is the only way to tell
them apart and we discard it. If Google retires the model id, the user is told
their key was rejected and rotates a perfectly good key. Suggest keeping 1008
terminal but classifying on the reason text, defaulting to
`.configurationRejected`.

**3. The `error` frame we parse doesn't exist.**
`terminalFailure(from:)` matches on `error.status`, but
`BidiGenerateContentServerMessage` has no `error` field. The union is exactly
setupComplete, serverContent, toolCall, toolCallCancellation, goAway,
sessionResumptionUpdate. So the quota, permission and not-found mappings are
unreachable and their tests pin a shape the server never sends. What actually
happens on quota exhaustion: the socket closes, we retry six times over ~61s,
and report "connection lost". https://ai.google.dev/api/live

**4. Gemini's failure modes fall into the "degrade" branch.**
`AppDelegate.swift:675` escalates a system-audio terminal failure only when
`provider == .openAI`. That was correct with two providers: Apple Speech emits
only `.appleSpeechUnavailable`, and degrading to mic-only is the right response
to a local analyzer failure. Gemini is the first provider that can emit
`.authenticationFailed`, `.quotaExceeded` and `.configurationRejected` on that
side, and they all now hit the degrade path instead. Concretely, a bad key
closes both sockets with 1008 at once from two delegate queues; if the
system-audio side lands first the user gets a system-audio degradation notice
that has nothing to do with system audio, written into Activity, moments before
the session fails anyway. A wrong key is the most common first-run mistake, so
it's a coin flip on first run. The condition wants to key on the reason rather
than the provider.

**5. Are the language codes right? I couldn't confirm either way.**
We send `setup.inputAudioTranscription.languageCodes` as `["en"]` for English
and `["zh-cn"]` for Mandarin. The documented supported values are `en-US` /
`en-GB` / `en-IN` and `cmn-Hans-CN`. These values come from
`TranscriptionLanguage.multipleHint`, whose own comment describes OpenAI's
contract ("newer transcription models accept a `languages` list"), reused for
Gemini on the strength of the assertion at `TranscriptionLanguage.swift:8` that
Gemini expects the same format. The docs don't say what an unlisted or shorter
code does, so I can't call this broken. `en` is valid BCP-47 and may normalize;
`zh-cn` to `cmn-Hans-CN` is the bigger gap. You have a live probe from a4acff2.
Could you send a setup with `["zh-cn"]` and see whether it's accepted, ignored,
or closed? If it's ignored, the Mandarin setting is a no-op today.

**6. Drop the intermediate docs; the wiki is the source of truth.**
`docs/superpowers/` is 2126 lines of spec and plan, and it's already stale: it
says "not yet implemented", `audioFrame(base64PCM:)` has since gained
`sampleRate`, and it claims Realtime and AppleSpeech read an injected format
when both still hardcode `pcm16Mono24k`. Intermediate design and planning docs
shouldn't land in the repo at all. Anything in there still worth keeping (the
rejected alternatives, and the reason the OpenAI socket lifecycle wasn't
extracted) should be folded into the relevant wiki page beside the design it
explains, and the rest dropped. `.superpowers/` is already gitignored; this is
the same thing under a committed name.

ironman-mark-I:
All six addressed — `ef7d944`, `fefb1ac`, `94a6773`. 958 tests / 111 suites, kernel guard passing.

Two of them I could settle with the live probe, and one answer is worse than "unconfirmed", so taking those first.

---

**5 — language codes. Cannot be determined behaviourally, so I switched to the documented values anyway.**

Two probes. Setup acceptance, one socket per case:

```
[]                      -> ACCEPTED      ["zh-cn"]               -> ACCEPTED
["en"]                  -> ACCEPTED      ["cmn-Hans-CN"]         -> ACCEPTED
["en-US"]               -> ACCEPTED      ["xx-NONSENSE"]         -> ACCEPTED   <- control
```

The control is the finding: `languageCodes` is not validated at setup, so acceptance proves nothing. So I tried a behavioural discriminator — English speech, four hint sets, comparing finals:

```
[]              -> "The index makes the query faster because it avoids a full table scan."
["zh-cn"]       -> identical
["cmn-Hans-CN"] -> identical     <- the DOCUMENTED Mandarin hint does not force Chinese either
["en-US"]       -> identical
```

So the hint is a soft bias, not a constraint, and no test distinguishes a valid code from garbage on clear speech. Your "I can't call this broken" is the correct verdict and I can't improve on it with evidence. I switched Gemini to `en-US` / `cmn-Hans-CN` regardless, since sending documented values costs nothing and removes an unverifiable assumption — but the comment says plainly that it is risk reduction rather than a proven fix, so nobody later reads it as one. It goes through a new `geminiHint`; `multipleHint` is untouched, because it is shared with OpenAI and changing it would alter the primary path's payload for no reason.

**3 — confirmed independently.** My bad-key probe produced a close with code 1008 and no `error` frame at all, matching your reading of the message union. Removed `terminalFailure(from:)`, its call site, and its tests — they were pinning a shape the server cannot produce, so they were fiction rather than coverage.

---

**1 — drain-then-rotate, plus the residual limitation documented.** `goAway` is advance warning, so the adapter now uses it: stop feeding the expiring socket (audio queues on the existing bounded FIFO), give the in-flight utterance a bounded grace period to land its final on the still-open socket, then rotate. The bound is `min(graceTimeout, server timeLeft)` — waiting past Google's own stated deadline would just trade a self-inflicted timeout for the same lost final. If the final never arrives, rotation proceeds anyway; an unbounded drain would hang the stream, which is worse than the garbled line. A rotation deliberately spends no reconnect budget and never reports `.failed`, mirroring how `rotating` already works for OpenAI's session cap. `wiki/status.md` records the residual case (a very long utterance can still straddle the boundary) and names `sessionResumptionUpdate` as the complete fix, left as a follow-up.

**2 — 1008 now classifies on the reason, defaulting to `.configurationRejected`.** Kept terminal either way, since no retry helps. Only an authentication-shaped reason upgrades to `.authenticationFailed`; anything unrecognised defaults to config-rejected, so a retired model id can never tell someone to rotate a good key. The wording is not a documented contract, so the default is the safe side and the comment says so.

**4 — the condition now keys on the reason, not the provider.** You were right that this wanted restructuring rather than another provider case: it was never about which provider, only about whether a failure affects one stream or both. The split lives on `TranscriptionFailureReason.affectsEveryStream` with a `CaseIterable` test over every case, so a future reason forces a deliberate escalate-or-degrade decision instead of silently defaulting to degrade. `.connectionLost` stays non-escalating on purpose — a dropped system-audio socket is a transport blip specific to that stream, not evidence the account is broken.

**6 — dropped.** 2126 lines gone. The rejected alternatives and the reason the OpenAI socket lifecycle was not extracted are now in `wiki/architecture.md` beside the design they explain; no in-repo references remain.

---

Two things I would rather flag than leave for you to find:

- The grace timeout is a 5s init default rather than threaded through `Config` like its sibling timeouts, and there is no live-verified number behind it — the rotation path is not in the smoke checklist yet, since triggering it means holding a session open for ~10 minutes.
- A rotation briefly surfaces as `.reconnecting(attempt: 1)`, the same state a real recovery uses. That mirrors OpenAI's existing behaviour rather than adding a state to distinguish expected churn from fault recovery.

ironman-mark-I:
Follow-up in `f88c9ba` — a 62-minute live session turned up a bug in the drain I added for finding 1.

All six rotations fired cleanly, but every one took **exactly 5.001s** and reported `grace period elapsed`. The early-rotate path never fired once. Cause: `beginDrain` checked only the socket lease, never whether there was an utterance to wait *for* —

```swift
let shouldDrain = self.task === task && self.generation == socketGeneration
    && !stopped && !isReconnecting && !rotating      // nothing about recognitionInFlight
```

— so a `goAway` always armed the full grace period, protecting a final that was never coming. That stalls new audio for up to 5s every ~9 minutes on both sockets: buffered instead of streamed, delaying the transcript and the coaching gated on it. My spec's fault; I wrote "give the in-flight utterance a bounded grace period" without adding *only if there is one*.

Now split into two conditions — no utterance in flight rotates immediately, otherwise the bounded deadline arms as before. The measured evidence is in the comment so the guard doesn't get removed later as a micro-optimisation.

Nothing was lost in that session: no speech landed in any of the six drain windows, so the stall never bit. It was luck, not correctness.

**Everything else held up under an hour of real load.** Zero errors, terminal failures, buffer overflows, key leaks, or continuity-witness stalls; 86 finalized lines; socket handoff averaged 372 ms. The binary-frame, settlement, lease-revalidation, and failure-classification fixes all behaved.

958 tests / 111 suites, kernel guard passing.

JINGBANZ (approved):

PR #269, opened 2026-09-08
feat(overlay): add a header to the Overlay Box and draw its resize affordance

JINGBANZ:
The Overlay Box carried no chrome of its own. Clearing it meant finding "Clear Overlay" in the menu bar, getting it out of the way meant switching it off in Settings, and its edges gave no sign they could be dragged. It now has a header, and it draws and owns its resize affordance.

Verified live in a real session: header, collapse, clear, header scaling, edge and corner resizing with Jarvis in the background, and drag-to-move.

## Header

| | |
|---|---|
| **Collapse**, left | `chevron.down` while open, `chevron.right` while collapsed, the macOS disclosure convention. Rolls the panel down to the header and back. |
| **Jarvis**, middle | Centred across the full width, so the name stays put whether or not the clear button is on screen. |
| **Clear**, right | The same `eraser` symbol the menu item used. Appears only when there is something to erase. |

- **Collapse is lossless.** The width and the height the user dragged to both survive the round trip, the collapsed height is never reported through `onSizeChanged`, and the drag floor meets the drag ceiling while collapsed so a vertical drag cannot stretch a box with nothing in it. The horizontal edges stay live, and a width drag then persists the *expanded* height.
- **Collapse does not persist.** Start rolls a collapsed box open again: it belongs to the conversation it was made during, not to a preference.
- **Clear replaces the menu item.** `MenuBarController.onClearOverlayBox` and the "Clear Overlay" item are gone.
- **A cleared box is blank**, as before.

Nothing about the header is a literal. Every dimension comes off the box's content height and is then clamped (`OverlayBoxChrome`), so the strip stays aimable at the 140 pt floor and still reads as chrome on a box dragged to fill a display:

| | |
|---|---|
| header height | `clamp(26, round(boxHeight * 0.075), 44)`, also the collapsed height of the panel |
| title | `clamp(11, round(header * 0.44), 19)` |
| button | `header - 8` square, radius `button * 0.22` |
| icon | `round(header * 0.5)` |
| side inset | `round(header * 0.17)` |

The title tracks the header rather than the Settings text-size slider, so that slider keeps meaning "how big are the tips".

## Resize affordance, and why it is not a cursor

A borderless window advertises no resize affordance. The obvious fix is a resize cursor, and the first pass did that. It only worked while the Settings window happened to be open, because Settings switches the app to `.regular` and activates it (`SettingsWindow.swift:28,31`) and back to `.accessory` on close. **macOS does not let an inactive application change the cursor**, and Jarvis is a background app for all of a session, so the affordance was missing exactly when it was needed.

A probe over a real session settled which half was broken:

```
12:38:48.109 entered at (623,379) zone=right       appActive=false policy=1
12:38:49.549 moved   at (620,273) zone=right       appActive=false policy=1
12:38:51.034 entered at (613,0)   zone=bottomRight appActive=false policy=1
```

92 events, all eight zones detected, every one with `appActive=false`. The `.activeAlways` tracking area works while inactive. Only `NSCursor.set` is ignored.

There is no public API for this. Apple DTS recommends exactly the tracking-area approach used here and the reporter confirms it does not work ([thread 738051](https://developer.apple.com/forums/thread/738051)); Apple's own `screencapture` gets its crosshair from private SkyLight calls (`CGSSetSystemDefinedCursor`, `CGSRegisterCursorWithImages`), which is not a dependency worth taking for a cursor. So the panel draws the affordance itself, the way Picture-in-Picture reveals its own controls on hover.

| | |
|---|---|
| Pointer on an edge | That run lights at 2.5 pt, 90% white. Which run lights is the direction. |
| Pointer on a corner | An L across both axes, arcing through the panel's own 12 pt radius with a 14 pt arm down each edge. Reachable by sliding along either edge. |
| Collapsed | Each side lights its whole rounded cap. A collapsed box is barely taller than its corner radius, so a straight run would be a stub. |
| Interior | Nothing. The log stays undisturbed while the pointer crosses it. |

Fades are 120 ms, and zero under Reduce Motion. The run is inset by its half-width plus a point, because the panel clips to its rounded corners and would otherwise eat the outer half of the stroke.

## Owning the drag, and one AppKit trap

Two things used to claim a drag on an edge: `isMovableByWindowBackground` moved the window, and AppKit's borderless edge-resize resized it. AppKit's resize region is far thinner than the 6 pt / 14 pt zones the affordance lights, so most of the lit area moved the box and only the innermost sliver resized it. The same edge did different things on consecutive tries.

The affordance now owns the drag through a pure `resizedFrame(draggingTo:from:startFrame:zone:minSize:maxSize:)`: the dragged edge follows the pointer, the opposite edge is anchored, and the window's own `minSize`/`maxSize` stay the single source of the range. AppKit no longer drives these resizes, so `viewDidEndLiveResize` does not fire for them and `onResizeFinished` reports the size instead; without that a dragged size would have stopped persisting.

The trap, worth knowing about: **`mouseDownCanMoveWindow` is not consulted per hit test. AppKit applies a `false` to the view's entire frame.** The affordance is a full-size sheet over the box, so refusing the drag there declared the whole box undraggable, however carefully its `hitTest` returned nil away from the edges. A hit-test-based test passed the whole time the app was frozen, because it tested the wrong mechanism.

So the refusing region is now the resizable ring itself: four `ResizeGripView` strips, one per edge and `edgeReach` thick, are the only views that refuse a window drag. The full-size sheet returns true and goes back to drawing and tracking; `hitTest` hands the press to whichever grip covers the point, and the grip forwards it to the parent's resize logic. Collapsed, the top and bottom strips shrink to nothing, so those edges neither resize nor block a drag. A test now encodes the rule: every drag-blocking frame must be no thicker than an edge strip.

One known trade-off: the header buttons sit 4 to 7 pt from the box edge, so their outermost couple of points resize rather than click.

## Runtime safety

Unchanged. No new window, no activation, no new presentation path. The header and the runs are drawn inside the panel that is already capture-excluded, so a screen share never sees them. `check-ghost-mode.sh` passes.

## Verification

`swift build && ./scripts/run-tests.sh` passes with zero new warnings. 50 Overlay Box tests cover the chrome maths across the full height range, the eight zones, what draws and when, the resize geometry and its clamping, which regions may refuse a window drag, the collapse round trip, the size-reporting guards, and the clear button's visibility.

## Summary by CodeRabbit

- **New Features**
  - Added a responsive Overlay Box header with collapse/expand, clear, and draggable title controls.
  - Added adaptive header sizing and highlighted edge and corner resize zones.
  - Collapsed boxes retain their position and support width resizing.

- **Bug Fixes**
  - Improved resizing behavior, size limits, top-edge anchoring, and Settings preview handling.
  - The clear control hides when no content is present and preserves live logs during previews.

- **Documentation**
  - Updated Overlay Box architecture and resize guidance.

- **Removed**
  - Removed the “Clear Overlay” menu bar item.


JINGBANZ:
## Interactive mock

An interactive mock of what this PR ships, for anyone reviewing the visual side:

**https://claude.ai/code/artifact/f1cf726e-5538-43ee-98b4-0bdc8fec8114**

The panel on it is live. Put the pointer on an edge or a corner to light it, drag to resize and watch the header rescale with the box, click the chevron to roll it up and the eraser to empty it. The switcher covers the open, collapsed and just-cleared states.

It is built to the same numbers as the code, not eyeballed: the box fill is `NSColor(white: 0.10, alpha: 0.45)`, the chrome comes from the `OverlayBoxChrome` formulas, and the zone logic is a port of `zone(at:in:allowsVerticalResize:)` with the same 6 pt edge and 14 pt corner reaches, y-axis flipped for the DOM. Two deliberate fidelity choices: there is no `backdrop-filter`, because the real panel is a flat translucent fill rather than an `NSVisualEffectView`, and the pointer stays an arrow throughout, because that is what you actually get.

That last point is the reason for the whole design, and the page carries the evidence: macOS does not let an inactive application change the cursor, and Jarvis is a background app for all of a session. The probe output on the page is from a real session.

EdHasNoLife (dismissed):

PR #271, opened 2026-09-08
fix(prompts): reduce music hallucinations and guide reply language

EdHasNoLife:
Background music can produce phantom transcripts and cause the coach to switch reply languages. This change adds language-independent transcription instructions to ignore music, singing, and background noise and avoid inventing speech. The context reaches both microphone and system-audio GPT Transcribe/GPT Live sessions; GPT-4o and Apple Speech retain their existing provider capabilities.

The coach follows the user's conversational language unless one language is explicitly selected, in which case that language guides replies. OpenAI and Gemini use their respective language selection; Apple Speech uses its selected locale. The reply instruction is resolved at Start and shared by per-turn and persistent CLI prompts, including subsequent brain-client refreshes. Automatic and multiple selections do not impose a fixed reply language.

This is prompt guidance only: Automatic remains unrestricted, bilingual input stays available, and model output is delivered without a second language classifier. No provider-metadata rejection, text filter, new suppression outcome, or audio-classifier behavior is included.

Validation:
- `swift build` passed after merging current main and resolving interview-format/Gemini integration conflicts.
- The exact Gate command (`swift build && ./scripts/run-tests.sh`) exits 0, but the combined test process ends inside the unchanged Overlay Box suite without a completion summary on this host. A serial run has the same limitation.
- Full-suite verification is therefore also run in isolated processes: all tests outside `OverlayBoxPanelTests` pass (982 tests in 116 suites); all 41 Overlay Box tests pass individually with completion summaries, for 1,023 verified tests total. No test implementation was changed or omitted.
- Focused prompt/configuration tests cover unrestricted Automatic payloads, both recording contexts, selected reply-language wiring (including Gemini), and preservation of bilingual context/output. Independent review found no actionable findings.
- Live music/audio and model compliance are not established by these tests; the live checklist is documented in `wiki/status.md`.

Language-independent root-cause investigation is tracked separately in #280. Prompt instructions alone cannot guarantee the absence of music hallucinations.

## Summary by CodeRabbit

* **New Features**
  * Coaching replies now follow the selected expected language when exactly one language is configured.
  * Without a single selected language, coaching continues to follow the conversational language.
  * Transcription guidance now ignores music, singing, and background noise instead of inventing speech.
* **Documentation**
  * Updated settings and architecture documentation to explain language-aware coaching and noise handling.
* **Tests**
  * Added coverage for language selection, multilingual transcripts, and background-noise transcription behavior.


JINGBANZ:
Thanks for digging into this, the Russian-from-music failure is real and worth fixing. But I don't want to merge the rejection layers, for a few reasons.

**Bilingual interviews are common.** With Mandarin selected, an English sentence is now dropped from the transcript, and the coach never sees it. With English selected, the reverse. Before this PR the selection was a hint and both languages came through. That's a regression in a core use case.

**Language is a proxy for the real problem.** The actual failure is the transcriber hallucinating text from non-speech audio. Rejecting by language catches Russian lyrics but also catches a real English colleague, and misses music hallucinated as plausible English (as the PR notes).

**Dropping model output on a second model's guess isn't a practice I want here.** The coach prompt used to be "the only place response behavior is governed (no code-side guardrail)," and this PR removes that line. Our rule is "build the harness, not model or OS intelligence." I'd rather keep the model in charge of what it says and how it says it. The NaturalLanguage detector also has no measured false-positive rate; 11 fixtures is a smoke test, and code-switched sentences ("我们用 Kafka 做 message queue") are exactly where it will misfire.

**Automatic changed meaning.** It went from "any language" to "English or Mandarin only." Anyone speaking another language now has a broken product, and adding a language means editing an enum plus the filter.

**What I'd like to merge from this PR:**
- Transcription prompt: "ignore music, singing, background noise, don't invent speech." This is language-independent and can ship on its own.
- Coach prompt: reply in the user's language, or the selected language if one is set. This alone fixes "Jarvis answered in Russian."

**What I'd like dropped:**
- Provider-metadata rejection, the `ConversationLanguageFilter`, the `suppressedLanguage` outcome, and the Automatic-means-en+zh change.

**Root cause, as a follow-up issue:** handle hallucination at the audio stage or with a confidence signal, both language-agnostic. Options: an on-device speech/music classifier (Apple SoundAnalysis) gating music-only segments before they reach the transcriber, or transcription logprobs from the Realtime API as a hallucination gate.

Can you split this into the prompt-only change? Happy to review that quickly.

EdHasNoLife:
Verified the current PR matches the requested prompt-only scope (877e492, now integrated with current main in 44b0d47). Provider-metadata rejection, ConversationLanguageFilter, suppressedLanguage, and the English/Mandarin-only Automatic restriction are absent. Bilingual input and model output remain intact; selected language only guides the prompt. The non-speech instructions are language-independent, and the audio-stage/confidence investigation is tracked in #280. The merge retains explanation/code guidance alongside the language instruction. Build and 29 focused prompt/language/shortcut tests passed; the full CI Gate is rerunning.

EdHasNoLife:
@coderabbitai review

EdHasNoLife:
Addressed the documentation coverage feedback in 4033123. Added the transcription-context API contract and documented the provider-resolution fixture and regression invariants: Automatic/bilingual selections preserve conversational choice, reply guidance never filters bilingual input or model output, and both provider payloads discourage non-speech hallucinations without adding a language restriction.

Validation: all 3 CoachReplyLanguageTests pass. The full Gate ran: build and static guards passed, and the runner exited 0, but again stopped without its full-suite completion summary. Complete CI validation is required for this documentation-only head. CodeRabbit must review this exact head before the warning is marked resolved; its current review cooldown is respected.

PR #360, opened 2026-09-17
fix(coach): restore placement headers above code hints

EdHasNoLife:
## Screenshot

Native AppKit detail view rendered from the PR branch with the example snippet. This demonstrates the restored placement header and inline code anchor; it is a rendering fixture, not a live model response.

Restored placement header above the code hint

Restore the short placement header above coding snippets, such as “Inside your `for right, ch in enumerate(s):` loop.” The coding skill now puts that context immediately before its fenced block in `detail`, so it stays with the code instead of appearing only in the transient hint. When code is not visible, the header states the placement assumption.

The general speak guidance defers to loaded skills without naming their content. Placement lives only in the detail header; hint lines carry the coaching. Neither code nor its header requires an explanation. This uses the existing Markdown renderer and keeps the current tool schema and overlay layout.

Validation: `swift build && ./scripts/run-tests.sh`; focused `DetailLayoutTests`, including the screenshot-shaped header example, visibility above the block, inline code styling, and unchanged snippet text. Live model output and the installed app were not exercised.

## Summary by CodeRabbit

- **Improvements**
  - Coaching details now provide supporting content requested by the active skill or needed to explain an answer.
  - Coding guidance now places a concise header immediately above each code block, using visible names and inline code references.
  - When code is not visible, guidance identifies the first relevant component and briefly states placement assumptions.

- **Documentation**
  - Architecture documentation now reflects the updated detail guidance and placement-header rules.

## End-to-end regression coverage

Scenario C now includes C33 for both ordinary hint presses. It checks the delivered Activity response for prose immediately before an accepted code block with a nonempty inline code anchor. Missing headers, placement only in hint lines, headers after the code, and absent inline anchors fail the check. It does not lock model wording or evaluate whether the named location is semantically correct.

Validation: the live test target compiled. `./scripts/run-live-tests.sh C` was attempted but preflight stopped before launch because the saved Gemini API key is missing (the existing harness checks all provider credentials even for Scenario C). A live pass remains unverified; rerun that command after saving the key in Settings → Connections. The offline Gate was run separately.

## Conflict-resolution validation

Current main is incorporated. The placement-header check is C33 so main’s C31/C32 preparation checks remain distinct. CI at `9aa26a8` passed 1,480 tests (20.883s): https://github.com/JINGBANZ/jarvis/actions/runs/35485009193. Local builds passed; the local full runner hit existing timing failures and did not print a final summary, so full-suite evidence comes from CI.


wiki/architecture.md
```diff
@@ -316,6 +316,11 @@ Show code for every implementation step. The pairing rules and exceptions live i
 [`coding` skill](../Sources/JarvisCore/Resources/Skills/coding/SKILL.md); the core keeps no second
 copy of that domain policy.
 
+Code details carry a short placement header above the fenced block, with inline code styling for
+visible anchors, so the location stays with the code when the hint scrolls away. This contextual
+header is allowed independently of explanation paragraphs; without visible code it states the
+placement assumption instead of inventing an editor location.
```
JINGBANZ, line 322: This restates the coding skill's rules two sentences after the page says the core keeps no second copy of that domain policy. The section opener above still defines detail as a code block, a diagram, or explanation paragraphs, so the page now contradicts itself. `wiki/AGENTS.md`: document each fact on one page and link elsewhere.

Please remove the paragraph. If the architecture page should mention that headers exist, a one-line pointer to the skill is enough.

Sources/JarvisCore/Coach/Tools/Speak.swift
```diff
@@ -55,7 +55,8 @@ private let detailGuidance = """
     The lines are the coaching. Keep the hint and its brief rationale there. Include detail when a
     loaded skill calls for supporting content or an explanation is warranted; otherwise leave it null.
     detail is Markdown shown under the hint in the persistent box. Use it for what a line cannot
-    hold: a code block or a diagram a loaded skill asked for. Write paragraphs there only when the
+    hold: a code block or a diagram a loaded skill asked for, with any short contextual header that
+    skill requires. Write explanation paragraphs there only when the
```
JINGBANZ, line 59: This guidance is in every session's prompt, so a rule about code headers is paid for by sessions that never load the coding skill, and each new kind of skill content would need another core edit. The core should only defer to the loaded skill, without naming what it asks for. The header rule then lives in the coding skill only.

```suggestion
    hold and a loaded skill asks for. Beyond that, write prose there only when the
```

Sources/JarvisCore/Resources/Skills/coding/SKILL.md
```diff
@@ -24,10 +24,13 @@ The block must implement that specific hint, not an unrelated step or an earlier
...
 merely to supply code; stay silent during healthy progress as usual.
 
-Put one fenced block in detail, tagged with its language (```python). Keep the hint itself the
-coaching: say in the lines where the block goes, using names visible on screen rather than editor
-line numbers. Code does not require a longer explanation; include explanation prose only when
-the explanation guidance calls for it.
+Put one fenced block in detail, tagged with its language (```python). Immediately before the block,
+include a short placement header saying where it belongs, using names visible on screen rather
+than editor line numbers. For example: Inside your `for right, ch in enumerate(s):` loop.
+Use inline backticks for code anchors. Keep this header with the code in detail, even when the
```
JINGBANZ, line 30: Pick one home for the where-it-goes sentence. With this clause, placement gets stated twice in a reply that has three short lines to spend. Drop "even when the hint lines also mention placement" and say the lines carry the coaching and do not repeat where the code goes. For example:

> The header, not the hint lines, says where the code goes; keep the lines for the coaching.

Sources/JarvisCore/Resources/Skills/coding/SKILL.md
```diff
@@ -24,10 +24,13 @@ The block must implement that specific hint, not an unrelated step or an earlier
...
+hint lines also mention placement. Keep the hint itself the coaching. A placement header does not
+require a longer explanation; include explanation prose only when the explanation guidance calls
+for it.
```
JINGBANZ, line 33: Before this PR the sentence said code itself needs no explanation. Narrowed to the header, the skill no longer says a code block is complete on its own, and a model can decide header plus code deserves a paragraph nobody asked for.

The sentence also points at "the explanation guidance", a section that was folded into the speak tool's Detail section in #338 and no longer exists under that name. The skill does not need to point at the core rule at all; it applies regardless. Suggested:

> Neither the code nor its header needs an explanation.

JINGBANZ (changes requested):
Four changes before this merges. Details are inline.

1. Remove the wiki paragraph; a one-line pointer to the skill at most.
2. Keep the speak guidance generic: defer to the loaded skill without naming what it asks for.
3. Give the placement header one home: the detail, not the hint lines too.
4. Keep the rule that code needs no explanation, and drop the stale cross-reference.

Tests/JarvisOverlayTests/DetailLayoutTests.swift
```diff
@@ -5,6 +5,36 @@ import Testing
...
+        view.show(detail, stamp: "10:30:00", position: (0, 1), isHeld: false, isRolled: false,
+                  fontSize: 18)
+        view.layoutSubtreeIfNeeded()
+        let scroll = try #require(view.subviews.compactMap { $0 as? NSScrollView }.first)
+        let document = try #require(scroll.documentView)
+        let texts = document.subviews.compactMap { $0 as? NSTextView }
+        let header = try #require(texts.first { $0.accessibilityLabel() == "Detail" })
+        let code = try #require(texts.first { $0.accessibilityLabel() == "Code block" })
+        #expect(!header.isHidden)
+        #expect(header.string == "Inside your for right, ch in enumerate(s): loop.")
```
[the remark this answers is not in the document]
EdHasNoLife: Keeping the exact-text assertion because this test supplies a fixed Markdown fixture directly to ReplyDetail; it never invokes the model or reads the coding skill. The assertion checks preservation of the surrounding prose while stripping Markdown delimiters, not generated wording. Changing prompt copy cannot break it. The separate anchor/font, geometry, and unchanged-code assertions remain in place.

Review follow-up: 77b88fc. Validation: 16 focused tests passed; `swift build && ./scripts/run-tests.sh` exited 0.

wiki/architecture.md
```diff
@@ -316,6 +316,11 @@ Show code for every implementation step. The pairing rules and exceptions live i
...
+
```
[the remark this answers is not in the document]
EdHasNoLife: Removed the duplicated paragraph. The existing coding-skill link now mentions placement headers without restating their rules, and the section opener describes generic skill-requested supporting content.

Review follow-up: 77b88fc. Validation: 16 focused tests passed; `swift build && ./scripts/run-tests.sh` exited 0.

wiki/architecture.md
EdHasNoLife: Removed the duplicated paragraph and kept placement headers in the existing skill cross-reference. Also made the section opener generic so it does not exclude skill-requested contextual content.

Review follow-up: 77b88fc. Validation: 16 focused tests passed; `swift build && ./scripts/run-tests.sh` exited 0.

Sources/JarvisCore/Coach/Tools/Speak.swift
EdHasNoLife: The detail guidance now defers to whatever a loaded skill asks for, without naming code, diagrams, or headers. Additional prose still follows the existing explanation criteria.

Review follow-up: 77b88fc. Validation: 16 focused tests passed; `swift build && ./scripts/run-tests.sh` exited 0.

Sources/JarvisCore/Resources/Skills/coding/SKILL.md
EdHasNoLife: Placement now has one home: the header in detail. The coding skill explicitly keeps hint lines for coaching without repeating placement.

Review follow-up: 77b88fc. Validation: 16 focused tests passed; `swift build && ./scripts/run-tests.sh` exited 0.

Sources/JarvisCore/Resources/Skills/coding/SKILL.md
EdHasNoLife: Replaced the narrowed rule and stale cross-reference with: “Neither the code nor its header needs an explanation.”

Review follow-up: 77b88fc. Validation: 16 focused tests passed; `swift build && ./scripts/run-tests.sh` exited 0.

JINGBANZ (dismissed):

JINGBANZ (approved):

PR #382, opened 2026-09-19
test(coach): cover malformed speak recovery boundaries

EdHasNoLife:
A malformed `speak` response can contain a string-valued `lines` field with embedded parameter markup and no `detail` field. Protect the existing recovery behavior with synthetic regression fixtures: only a valid replacement reaches the overlay, and repeated malformed calls fail at the existing seven-response attempt limit.

Cover the current Anthropic Messages adapter as well: invalid arguments remain available for schema feedback, and replay preserves the original tool input and call linkage. The tests cover both automatic and manual attempts. No runtime behavior or provider routing changes.

Validation:
- Full Gate passed in [CI](https://github.com/JINGBANZ/jarvis/actions/runs/35455278488) at `015f077`: `swift build` and `./scripts/run-tests.sh` completed on macOS 15. Final output: **1,476 tests passed after 17.125 seconds**.
- Targeted recovery and Messages suites: 30 tests passed.
- Mutation check: temporarily allowing an eighth response makes both attempt-budget regression cases fail; production limit restored.
- `swift build` passed. `swift build && ./scripts/run-tests.sh` was invoked, but the local test process exits 0 before the completion summary. Sequential full runs reproduce this during existing AppKit layout tests, so the local Gate is **incomplete**, not passing.
- Isolated non-AppKit inventory: all 1,331 tests passed with the final count verified. Several isolated overlay suites pass; navigation still exhibits the premature exit. Complete offline coverage is not established locally.
- LLDB caught `exit` called from Swift concurrency `_swift_exit` / `swift_task_asyncMainDrainQueue` during the existing layout tests, with AppKit animation threads active. No production or test-infrastructure workaround is included.

The complete CI Gate above clears readiness; the local runtime limitation remains documented separately.

The recorded failure was already present at the helper HTTP response boundary. The available evidence cannot distinguish model generation from the former proxy conversion; the current Messages adapter was introduced separately in #372. These tests do not claim a measured improvement in model reliability or latency.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for recovering from malformed speech tool arguments while preserving the original wire data.
  - Verified corrected replacement calls can complete successfully and display the appropriate hint for automatic and manual triggers.
  - Confirmed repeated malformed calls reach the call limit and produce a temporary response-stage error without rendering.


EdHasNoLife:
@coderabbitai review

JINGBANZ (approved):

